PR #1, opened 2017-03-09
[WIP] Use RxJS for Component loading & aborting

luigiplr:
By using a more Reactive Programming approach we can teardown our loader subscription & reduce any un-needed cpu overhead for components that have unmounted prior to fully loading.

An example of such a use-case would be in a chat client w/ some form of scroll virtualization. Where as the user scrolls up rapidly message-specific components may unmount before fully having the chance to load.

Another addition I'm thinking may be handy would be a way of tracking what components are already loading (LRU + Subjects?) so that duplicate components do not start two loaders in parallel.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,6 @@
     ]
   },
   "dependencies": {
-    "lodash.curry": "^4.1.1",
-    "lodash.curryright": "^4.1.1"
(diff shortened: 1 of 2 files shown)


thejameskyle:
Thanks, but I don't want to add a dependency on rxjs for this project.

PR #2, opened 2017-03-11
Update index.js: Add logging

tiaghoalves:
Just adding some logging to be more safe.


thejameskyle:
The silent error is intentional. `require()` will throw an error when the module hasn't been loaded client-side yet. This might be improved later on via splitting up the client-side and server-side sync importing. Waiting on a `require.id` feature to be built into Webpack for this.

PR #5, opened 2017-03-12
NPM module fix

yatharthk:
Currently, the `npm install`ed module doesn't work for two reasons:

1. absence of `lib`
2. `main` key in `package.json` points to `index.js` whereas it should point to `lib/index.js`

PR fixes the issue.


thejameskyle:
lol thx

Issue #8, opened 2017-03-14
Curious why `webpackRequireWeakId` is a function


kentcdodds:
Just curious.

thejameskyle:
`require.resolveWeak` does not exist in any environment except for Webpack. It would throw an error in Node if you did this:

```js
let LoadableMyComponent = Loadable({
  // ...
  webpackRequireWeakId: require.resolveWeak('./MyComponent'), // Error: require.resolveWeak is not a function
});
```

kentcdodds:
Ah, yeah, makes sense. Thanks!

PR #9, opened 2017-03-14
docs: add reference to my question

kentcdodds:
In case anyone else is curious :)

ref #8


Daniel15:
IMO you should add a summary of the contents of the issue, rather than just linking to it.  Perhaps an FAQ section?

kentcdodds:
Especially now that issues are disabled 😅

thejameskyle:
Added detail to the README

mohsinulhaq:
issues page disabled :/

PR #13, opened 2017-03-15
Propagate syntax error from server side rendering require

relekang:
Thanks for making this library :v:

While trying to add server side rendering to a project after I started to use react-loadable, I ran into a problem where it turned out that I tried to require an image and node wasn't to happy about that. I figured it would be a bit easier to catch if these errors were propagated to the error prop of the loading component. This is a PR trying to implement that.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -14,7 +14,11 @@ let tryRequire = (pathOrId: string | number) => {
   try {
     // $FlowIgnore
     return babelInterop(require(pathOrId));
-  } catch (err) {}
+  } catch (error) {
+    if (error.name === "SyntaxError") {
+      throw error;
+    }
+  }
   return null;
 };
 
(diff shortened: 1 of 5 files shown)


src/index.js
```diff
@@ -14,7 +14,11 @@ let tryRequire = (pathOrId: string | number) => {
   try {
     // $FlowIgnore
     return babelInterop(require(pathOrId));
-  } catch (err) {}
+  } catch (error) {
+    if (error.name === "SyntaxError") {
```
thejameskyle: Maybe instead we can exclusively catch only the exact errors which will be throw by node/webpack/browserify/etc when `require()`-ing something invalid.

src/index.js
```diff
@@ -43,11 +47,21 @@ export default function Loadable<Props: {}, Err: Error>(opts: Options) {
   let outsideError = null;
 
   if (serverSideRequirePath) {
-    outsideComponent = tryRequire(serverSideRequirePath);
+    try {
+      outsideComponent = tryRequire(serverSideRequirePath);
+    } catch (error) {
+      isLoading = false;
+      outsideError = error;
```
thejameskyle: We should probably `console.error` these or something.

src/index.js
relekang: Not entirely sure I follow. Do you mean adding more cases or changing how we check what kind of error it is?

thejameskyle:

PR #14, opened 2017-03-15
fix webpack integration

### LINKED ISSUE #10: Warning critical dependency
First of all James thanks for publishing this cool project 🙌 , i have tried and its working really good in SSR too, but i'm getting this warning in console and i don't know yet why! if anyone else is getting please let me know and if you did get rid of it.

```
WARNING in ./~/react-loadable/lib/index.js
31:24-41 Critical dependency: the request of a dependency is an expression
```

kentcdodds:
❗️ This is actually a much bigger deal than originally thought. Check out [this repo](https://github.com/kentcdodds/react-loadable-webpack-bug) for more info. Here's the README (saved you a click):

---

This demonstrates a bug in the way that react-loadable intracts with webpack. If you pop open the `index.js` you'll see it's simply importing `react-loadable`. Check out the `bundle.js` file (which is generated by webpack) and you'll see some interesting things in there:

1. [`webpackEmptyContext`](https://github.com/kentcdodds/react-loadable-webpack-bug/blob/b2e7226491d17426f96e7f97d81d3566a154732a/bundle.js#L232-L244) - a module that webpack makes when you do `require(somethingDynamic)`.
2. [`require` replaced by IIFE that throws](https://github.com/kentcdodds/react-loadable-webpack-bug/blob/b2e7226491d17426f96e7f97d81d3566a154732a/bundle.js#L104)

**So, basically, the try/catch there will _always_ result in a throw with an ignored catch.**

Now, if you look at [this commit](https://github.com/kentcdodds/react-loadable-webpack-bug/commit/15898ccf8027993a40efc81fa9c72395f1a56525), it uses [my branch of react-loadable](https://github.com/kentcdodds/react-loadable/tree/pr/remove-webpack-warning) instead and the problem is fixed. Now:

1. `webpackEmptyContext` is replaced by a [`module` polyfill](https://github.com/kentcdodds/react-loadable-webpack-bug/blob/15898ccf8027993a40efc81fa9c72395f1a56525/bundle.js#L233-L260) (to account for `module.require`). This is actually dead code because `module.require` will never be used when running this with webpack anyway. :shrug:
2. `require` is now `requireFn` which will be assigned to `__webpack_require__` when running with webpack and simply `module.require` when running in node.

I've verified that my fix works wonderfully in my app. :) 

---

This uses `__webpack_require__` if `isWebpack` is true and `module.require` if not. This should work in both node and when bundled with webpack without causing issues.

Closes #10

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -5,6 +5,9 @@ type GenericComponent<Props> = Class<React.Component<{}, Props, mixed>>;
 type LoadedComponent<Props> = GenericComponent<Props>;
 type LoadingComponent = GenericComponent<{}>;
 
+let isWebpack = typeof __webpack_require__ !== "undefined";
+let requireFn = isWebpack ? __webpack_require__ : module.require.bind(module);
+
 let babelInterop = obj => {
   // $FlowIgnore
   return obj && obj.__esModule ? obj.default : obj;
@@ -13,13 +16,10 @@ let babelInterop = obj => {
 let tryRequire = (pathOrId: string | number) => {
   try {
     // $FlowIgnore
-    return babelInterop(require(pathOrId));
+    return babelInterop(requireFn(pathOrId));
   } catch (err) {}
   return null;
 };
-
-// $FlowIgnore
-let isWebpack = typeof __webpack_modules__ !== "undefined";
 
 type Options = {
   loader: () => Promise<LoadedComponent<Props>>,
@@ -42,12 +42,19 @@ export default function Loadable<Props: {}, Err: Error>(opts: Options) {
   let outsidePromise = null;
   let outsideError = null;
 
-  if (serverSideRequirePath) {
+  if (!isWebpack && serverSideRequirePath) {
     outsideComponent = tryRequire(serverSideRequirePath);
   }
 
   if (isWebpack && webpackRequireWeakId) {
-    outsideComponent = tryRequire(webpackRequireWeakId());
+    let weakId = webpackRequireWeakId();
+    if (__webpack_modules__[weakId]) {
+      // if it's not in webpack modules, we wont be able
+      // to load it. If we attempt to, we mess up webpack's
+      // internal state, so only tryRequire if it's already
+      // in webpack modules.
+      outsideComponent = tryRequire(weakId);
+    }
   }
 
   let load = () => {
--- a/flow-typed/webpack.js
+++ b/flow-typed/webpack.js
@@ -1,0 +1,2 @@
+declare function __webpack_require__(pathOrId: string | number): any;
+declare var __webpack_modules__: {};


garmjs:
This is great :), hopefully this get merged thanks @kentcdodds

kentcdodds:
Let's hold off for a second. I'm having some issues with this in my app... Debugging now

garmjs:
@kentcdodds thanks for letting me know, i didn't get chance to test it yet

kentcdodds:
Fixed! This _now_ works swimmingly in my app :) I promise

garmjs:
This was fast 🙏

src/index.js
```diff
@@ -1,26 +1,38 @@
...
   try {
+    if (
+      isWebpack &&
+      typeof pathOrId === "number" &&
+      !__webpack_modules__[pathOrId]
+    ) {
+      // if it's not in webpack modules, we wont be able
+      // to load it. If we attempt to, we mess up webpack's
+      // internal state, so we exit early
+      return null;
```
kentcdodds: Here's a bit of the `__webpack_require__` function:

```javascript

/******/ 	// The require function
/******/ 	function __webpack_require__(moduleId) {

/******/ 		// Check if module is in cache
/******/ 		if(installedModules[moduleId])
/******/ 			return installedModules[moduleId].exports;

/******/ 		// Create a new module (and put it into the cache)
/******/ 		var module = installedModules[moduleId] = {
/******/ 			i: moduleId,
/******/ 			l: false,
/******/ 			exports: {}
/******/ 		};

/******/ 		// Execute the module function
/******/ 		modules[moduleId].call(module.exports, module, module.exports, __webpack_require__);
// ... yada yada
```

So, from this we learn that `installedModules` gets initialized with the given `moduleId` when you call `__webpack_require` and then the next time you try to call `__webpack_require`, it exits early with what it is initialized with. So what was happening was we'd call it here, then my dynamic import would call it and it would exit early with that empty `export` object.

So we can protect against this by not calling the `requireFn` if `__webpack_modules__` (which I believe is just `installedModules` has not yet been initialized with the module).

src/index.js
```diff
@@ -1,26 +1,34 @@
...
   return obj && obj.__esModule ? obj.default : obj;
 };
 
 let tryRequire = (pathOrId: string | number) => {
   try {
+    if (isWebpack && !__webpack_modules__[pathOrId]) {
+      // if it's not in webpack modules, we wont be able
+      // to load it. If we attempt to, we mess up webpack's
+      // internal state, so we exit early
+      return null;
```
kentcdodds: Here's a part of `__webpack_require__`:

```javascript

/******/ 	// The require function
/******/ 	function __webpack_require__(moduleId) {

/******/ 		// Check if module is in cache
/******/ 		if(installedModules[moduleId])
/******/ 			return installedModules[moduleId].exports;

/******/ 		// Create a new module (and put it into the cache)
/******/ 		var module = installedModules[moduleId] = {
/******/ 			i: moduleId,
/******/ 			l: false,
/******/ 			exports: {}
/******/ 		};

/******/ 		// Execute the module function
/******/ 		modules[moduleId].call(module.exports, module, module.exports, __webpack_require__);
// ... yada yada
```

So what's happening here is if the given `moduleId` exists in `installedModules`, it exits early with that module's `exports`. If it doesn't, then an empty one is created and then the module function is called with `.call`. However, in our case, the module has yet to be loaded so the `.call` call throws an error. We catch and ignore that error which is fine, but then the next time we come around (when loading it asynchronously) we're going to exit early with that empty `exports` object that it was initialized with.

So to fix this, we check whether `__webpack_modules__` (which I think is just the same thing as `installedModules`) already has the module. If it doesn't, then we're going to skip calling `__webpack_require__` because it's going to fail anyway (and mess up the state of `installedModules`).

Because this is kinda confusing I thought it'd be good to include a comment explaining this briefly.

kentcdodds:
> This was fast 🙏

I'm working on something right now that uses this, so I was motivated :)

Thanks for giving it a look when you get around to it @thejameskyle :)

kentcdodds:
I went ahead and published `@kentcdodds/temp-react-loadable` (I DO NOT PLAN TO MAINTAIN THAT SO PLEASE DON'T DEPEND ON IT). I just published it so it's easier for people to test out. It's working great for me.

navgarcha:
@kentcdodds you legend. I was digging around this code to solve my https://github.com/thejameskyle/react-loadable/issues/12 issue and this looks to do just that. Now if i include the normally async chunk (that was used on the SSR) directly in a script tag on my page i no longer get a client-server mismatch! Just need that `flushServerSideRequires` [functionality](https://gist.github.com/thejameskyle/abecfe8ec2a7ce1e312a904527a31908) and we're golden!

Great work!

garmjs:
@navgarcha always bet on @kentcdodds 🎲

src/index.js
```diff
@@ -1,10 +1,14 @@
 // @flow
 import React from "react";
 
+declare function __webpack_require__(pathOrId: string | number): any;
```
thejameskyle: Can you just move this all to a `flow-typed/webpack.js` file?

src/index.js
```diff
@@ -1,26 +1,34 @@
...
+let requireFn = isWebpack ? __webpack_require__ : module.require;
+
 let babelInterop = obj => {
   // $FlowIgnore
   return obj && obj.__esModule ? obj.default : obj;
 };
 
 let tryRequire = (pathOrId: string | number) => {
   try {
+    if (isWebpack && !__webpack_modules__[pathOrId]) {
```
thejameskyle: Can you move this all down?

```diff
-  if (serverSideRequirePath) {
+  if (!isWebpack && serverSideRequirePath) {
     outsideComponent = tryRequire(serverSideRequirePath);
   }
 
   if (isWebpack && webpackRequireWeakId) {
     outsideComponent = tryRequire(webpackRequireWeakId());
   }
```

src/index.js
kentcdodds: Done :+1:

src/index.js
kentcdodds: I can move it down, but it'll look slightly different from what you have there.

thejameskyle: You only need the `__webpack_modules__[pathOrId]` because you're calling ` tryRequire(serverSideRequirePath);` right? If you don't call that with a `!isWebpack` I don't see why you'd need anything different

src/index.js
```diff
@@ -1,10 +1,14 @@
 // @flow
 import React from "react";
 
+declare function __webpack_require__(pathOrId: string | number): any;
 type GenericComponent<Props> = Class<React.Component<{}, Props, mixed>>;
 type LoadedComponent<Props> = GenericComponent<Props>;
 type LoadingComponent = GenericComponent<{}>;
 
+let isWebpack = typeof __webpack_require__ !== "undefined";
+let requireFn = isWebpack ? __webpack_require__ : module.require;
```
benjamn: I think you may need to bind that require method (`module.require.bind(module)`) in order to call it safely as a simple function below.

kentcdodds: Thanks for checking, but I've verified that this works fine in node:

```javascript
// temp.js
const requireFn = module.require;

console.log(requireFn('fs'));
console.log(requireFn('./package.json'));
console.log(requireFn('./lib/index'));
```

Are there other contexts where this could be a problem though?

src/index.js
thejameskyle: oh i see

kentcdodds:
I think this is mergable. I've verified that the changes work in my project as well :+1:

src/index.js
```diff
@@ -42,12 +42,19 @@ export default function Loadable<Props: {}, Err: Error>(opts: Options) {
...
 
-  if (serverSideRequirePath) {
+  if (!isWebpack && serverSideRequirePath) {
     outsideComponent = tryRequire(serverSideRequirePath);
   }
 
   if (isWebpack && webpackRequireWeakId) {
-    outsideComponent = tryRequire(webpackRequireWeakId());
+    let weakId = webpackRequireWeakId();
+    if (__webpack_modules__[weakId]) {
```
thejameskyle: I feel like you should guard this with a `typeof __webpack_modules__ === 'object'` check.

navgarcha: wouldn't the `isWebpack` above be sufficient? `__webpack_require__` and `__webpack_modules__ ` are both internals of a webpack runtime. Otherwise `let isWebpack = typeof __webpack_require__ !== "undefined" && typeof __webpack_modules__ !== "undefined";` might be neater?

kentcdodds: We could do that, though we're pretty much guaranteed that `__webpack_modules__` will be an object if `isWebpack` is true. I could also change `isWebpack` to be:

```javascript
let isWebpack = typeof __webpac_modules__ !== "undefined";
```

Which is what it was before, but it really makes no difference. Happy to do whatever you decide, but I think doing a `typeof` check is unnecessary here.

src/index.js
benjamn: I'm having trouble seeing how this could work, since `Module.prototype.require` refers to `this` expecting it to be the `module` object, and is never bound to that object (not even [here](https://github.com/nodejs/node/blob/5bfd13b81e38b60a7b9f346fbfcb216192cf0974/lib/module.js#L582), in case you had a similar hope).

Unless you have an explanation of how this works, I have to insist that it's dangerous to call `module.require` without knowing what `module.id` (the parent module identifier) is.

If it's not clear why this works (if it does), please just bind it to be safe!

kentcdodds: Sure thing, I've got no problem binding it :)

src/index.js
```diff
@@ -5,6 +5,9 @@ type GenericComponent<Props> = Class<React.Component<{}, Props, mixed>>;
 type LoadedComponent<Props> = GenericComponent<Props>;
 type LoadingComponent = GenericComponent<{}>;
 
+let isWebpack = typeof __webpack_require__ !== "undefined";
+let requireFn = isWebpack ? __webpack_require__ : module.require.bind(module);
```
benjamn: Thanks!

navgarcha:
@edorivai could you not just check if your rendering on the server and sync `require` instead of using `Loadable`? Something like: 
```
// src/components/Header/loadable.js
export default process.env.SERVER ? require('.') : Loadable({
    loader: () => import('.'),
    LoadingComponent: Loading,
    webpackRequireWeakId: () => require.resolveWeak('.')
});
```

edorivai:
Wouldn't that result in webpack bundling the component in the client bundle, instead of splitting it?

navgarcha:
No, you would use the Webpack define plugin to set `process.env.SERVER` accordingly and this will remove the redundant code branch

edorivai:
Right, makes sense. I'll give it a go, thanks!

edorivai:
Ah damn. [James' snippet](https://gist.github.com/thejameskyle/abecfe8ec2a7ce1e312a904527a31908) led me to believe that `flushServerSideRequires` was already implemented :stuck_out_tongue:. But `require()`-ing it on the server would make it impossible for `react-loadable` to know which components were loaded, right?

navgarcha:
Yea i asked this question also - it appears that `flushServerSideRequires` is still work in progress or 'we could do this' code. I tried digging around the source to implement this but havent had time to look into it properly.

ctrlplusb:
@edorivai @navgarcha yep, and this is the exact feature I am waiting on before making a switch from my `react-async-component` to this lib.

edorivai:
Aha, that clears things up. I've been messing around a bit, and there are quite a few hurdles to be taken. For example, the `() => import(...)` would work on the server if `import` could somehow resolve synchronously. Using serverSideRequirePath has it's own problems, especially when you bundle your serverside rendering code with webpack, because you need webpack-loader functionality for instance.

I tried forcing the use of `module.require` within `react-loadable`, but this doesn't work because:

1. The `serverSideRequirePath` doesn't match relative to my output bundle.
2. When I try to solve (1) by specifying `serverSideRequirePath` manually, and point to my source, Node complains about not understanding es2015, jsx etc.

I can imagine this all works seamlessly if you just run your source through babel, and get a babelified tree that mirrors your source tree, but otherwise it's kinda hard...

thejameskyle:
`flushServerSideRequires` isn't very hard to write, I'll put up a branch

thejameskyle:
See #29

PR #17, opened 2017-03-17
Preload returns the load Promise

### LINKED ISSUE #16: Return the promise from preload
Right now preloading doesn't return the internal `load()` promise, but it could be useful having it.

My use case is: I do server side rendering, the "route" chunks get loaded asynchronously, but I wait for the "above the fold" ones to get fully fetched prior to render the React app.
This way I don't lose server side rendered content while waiting for the chunks to arrive.

BTW: great job! ^_^

EnoahNetzach:
fixes #16

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -78,7 +78,7 @@ export default function Loadable<Props: {}, Err: Error>(opts: Options) {
     _mounted: boolean;
 
     static preload() {
-      load();
+      return load();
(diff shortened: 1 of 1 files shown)


thejameskyle:
Closing, see https://github.com/thejameskyle/react-loadable/issues/16#issuecomment-287485305

PR #21, opened 2017-03-22
expose promise in preload function

petrhanak:
It is surprising this preload function is not returning promise from `load`. This way we could trigger custom logic after preloading finished.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -78,7 +78,7 @@ export default function Loadable<Props: {}, Err: Error>(opts: Options) {
     _mounted: boolean;
 
     static preload() {
-      load();
+      return load();
     }
 
     state = {


thejameskyle:
There was the explicit goal of not allowing users to depend on the timing of `preload()`. The idea that you shouldn't be deferring any action until after it is preloaded. What is your goal here?

petrhanak:
Right now the library is only for loading from "nothing to display" to "content loaded". Using this (and maybe more advanced API). It would be availiable switching content e.g. routing. Flow Will look like: "some content loaded", "loading new content" (old content displayed with loading indicator), "new content loaded".

thejameskyle:
> Right now the library is only for loading from "nothing to display" to "content loaded"

That's not true, this library is explicitly for the purpose of rendering a loading component before new content is loaded. I suggest you look at https://github.com/thejameskyle/react-loadable-example

damienadermann:
Returning a promise here is very helpful for chunked requests with server side rendering. If I can't tell when the preload has completed I'm forced manually import the chunk modules, then call `preload` and add a random timeout. If I don't add call preload the loadable component will flash the loading before showing the chunked component which looks bad and breaks the checksum.

thejameskyle:
There are going to be better ways to solve that in the future once we have Fiber.

PR #22, opened 2017-03-23
Please enable issues / make LoadingComponent optional

aikar:
@thejameskyle 
Why is issues disabled? Please enable issues so people can report issues to you.

Additionally, the issue I am trying to raise: LoadingComponent should be optional, and default to null (nothing displayed).

This module currently throws errors when it's not supplied.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ Function returning promise returning a React component displayed on success.
 Resulting React component receives all the props passed to the generated
 component.
 
-#### `opts.LoadingComponent`
+#### `opts.LoadingComponent` (optional, defaults to null)
 
 React component displayed after `delay` until `loader()` succeeds. Also
 responsible for displaying errors.
--- a/src/index.js
+++ b/src/index.js
@@ -121,7 +121,7 @@ export default function Loadable<Props: {}, Err: Error>(opts: Options) {
       let { pastDelay, error, Component } = this.state;
 
       if (isLoading || error) {
-        return (
+        return (!LoadingComponent ? null :
           <LoadingComponent
             isLoading={isLoading}
             pastDelay={pastDelay}


thejameskyle:
Issues are closed because as this proves, people open up pull requests instead.

Why do you want the LoadingComponent to be optional?

aikar:
Not everyone is able to contribute the solution to their problem... Issues enable people to report issues that they wish for you to resolve.

and I want it optional so I don't have to pass anything to it every time I use it. On some components, you may not want to show any form of loading indicator ever., and creating noop function stubs for every case is tedius.

thejameskyle:
It's not my job to resolve issues for other people on these projects. They can pay me if they want that. 

In general I would recommend wrapping `Loadable` with your own HOC which prepopulates the `LoadingComponent` prop. That way you dont have to reconfigure it every time.

aikar:
I guess if that's how you want to run your project... I recommend noting this on the readme that you are not going to provide support. I looked first to see if their was an external issue tracker but did not see it.

It's clear that you have strong opinions on what this project should be, so I'm going to opt to maintain a fork instead.

thejameskyle:
I have not said no to this pull request. I'm just trying to understand it's use case.

I provide excellent support for my projects. If I had issues enabled we'd be having this conversation either way. Except you probably never would have opened a pull request (if I am to assume your behavior matches the behavior of 99% of people I interact with in open source).

aikar:
How does one ask you for support if they have an issue with the project, but don't have the solution and they are looking to hold a discussion on finding the problem or simply asking for clarity?

Many users are juniors and are intimidated by contributing to larger projects, especially considering this project is using Flow and they might not understand it.

As for my use, sometimes the content you're loading is smaller than the loading indicator. I just don't need the indicator for that.
Or you may have lots of loading components in 1 area and too many indicators will go into overload.

Stacked items also benefit of having no indicator and simply sliding in when it loads.

It's better to let the user decide how they want to handle loading indicators, and if they don't want one, then the project shouldn't care, and just show nothing until it loads.

thejameskyle:
Listen I maintain a lot of projects, I've spent thousands of hours over the last few years contributing to open source. It's been extremely bad for my mental health and I'm struggling to make it worthwhile for myself.

Issues are extremely draining, I get yelled all the time whether I'm doing a good job or not. I'm sick of it, I don't get anything out of it and I'm making the extremely opinionated decision to opt-out.

If you have a problem with that, fine, go right ahead and fork the project. I license everything that I do as openly as possible. My open source is MIT/ISC and my writing is Creative Commons.

But if you have a problem with me picking my mental health over supporting people who are often abusive towards me, well you can go fuck yourself.

---

As for this feature, I'm of the opinion that APIs should optimize for correct usage and make incorrect usage difficult. In order to not show anything, all you have to do is:

```js
Loading({
  loader: () => import("./foo"),
  LoadingComponent: () => null,
});
```

You are being explicit about the fact that you don't intend to render anything. It's very little work to do and is pretty intuitive. Maybe it's not obvious, so I'll update the docs to include a note about it.

I think that this is the right decision for this package, and so I'm going to make the decision to close this pull request.

PR #23, opened 2017-03-23
Return the promise from load to preload

aikar:
So that one can preload multiple components and wait until all of them are done with Promise.all()


aikar:
nevermind, I see you have recently denied this... I strongly disagree, so guess i'll have to maintain a fork.

thejameskyle:
> I strongly disagree, so guess i'll have to maintain a fork.

I love that aspect of open source!

PR #24, opened 2017-03-29
custom resolve logic

edvinerikson:
Adds a option to pass in a custom resolve function that can be used if the component is not the default exported from a module.


thejameskyle:
Awesome work, thanks!

edvinerikson:
Thanks for quick review and merge 💯

PR #26, opened 2017-04-01
Move webpack require to component lifecycle

richardscarrott:
I'm proposing to move the initial Webpack require from the HoC closure -> the HoC constructor to allow markup from SSR apps to be reused by preloading Webpack Chunks.

i.e.

```
// index.html
<script src="/main.js"></script>
<script src="/chunk.0.js"></script>
<script>
    window.render(); // Exposed by main.js to allow us to wait for chunk 0 to load before rendering.
</script>
```

Without this change the Loadable HoC would be called as soon as main.js is parsed and executed meaning chunk 0 won't have executed yet.

Tbh, this change wouldn't be necessary if Webpack allowed you to load chunks before main, perhaps by registering themselves on the global and having the main chunk read them in when it has loaded, but I don't *think* Webpack supports this.

Also, if you're interested in merging this I can add tests.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -47,17 +47,6 @@ export default function Loadable<Props: {}, Err: Error>(opts: Options) {
     outsideComponent = tryRequire(serverSideRequirePath);
   }
 
-  if (isWebpack && webpackRequireWeakId) {
-    let weakId = webpackRequireWeakId();
-    if (__webpack_modules__[weakId]) {
-      // if it's not in webpack modules, we wont be able
-      // to load it. If we attempt to, we mess up webpack's
-      // internal state, so only tryRequire if it's already
-      // in webpack modules.
-      outsideComponent = tryRequire(weakId);
-    }
-  }
-
   let load = () => {
     if (!outsidePromise) {
       isLoading = true;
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -82,11 +71,26 @@ export default function Loadable<Props: {}, Err: Error>(opts: Options) {
...
 
-    state = {
-      error: outsideError,
-      pastDelay: false,
-      Component: outsideComponent
-    };
+    constructor(props) {
+      super(props);
+
+      if (isWebpack && webpackRequireWeakId) {
```
thejameskyle: Could you guard this with another check to see if `outsideComponent` already exists?

thejameskyle:
> Tbh, this change wouldn't be necessary if Webpack allowed you to load chunks before main, perhaps by registering themselves on the global and having the main chunk read them in when it has loaded, but I don't think Webpack supports this.

It doesn't, I was bothering @TheLarkInn about it before because I was trying to do all of the same stuff you're describing.

faceyspacey:
The best and idiomatic I've come up with is to use 
```
new webpack.optimize.CommonsChunkPlugin({
      names: ['bootstrap'],
      filename: '[name].js',
      minChunks: Infinity,
    }),
```
***but with no entrypoint called "bootstrap" in which case webpack puts all the bootstrap code in the first chunk.*** Then you put your "dynamic" split chunks before `main`, and `main` can synchronously require them. 

Also note: if you have a `vendor` bundle, you don't need to have a separate `bootstrap` bundle as the vendor bundle will also contain the bootstrap code. For my apps though I always embed the bootstrap code directly into the page since it's small and so it doesn't require an additional request. So having an auto-generated bootstrap chunk is natural for my setup.

richardscarrott:
@faceyspacey I knew I had seen an entry chunk before which didn't immediately execute a module but couldn't think where -- very nice indeed -- and embedding into the html seems reasonable to save on an extra http request although maybe less of a concern if using http2.

I guess that makes this PR unnecessary unless @thejameskyle or @TheLarkin have any thoughts on each approach?

faceyspacey:
...now we just need a simple API that packages up the required dynamic chunks and puts them before `main` ;) ...and works for both babel and webpack, and addresses CSS. AND handles challenges for development where you can't mix the style loader with the extract text plugin. And various other mackamoles :).

I've been working on precisely this for a week now. PR should be submitted tomorrow.

thejameskyle:
I'm also working on some tooling that can track timing information about when dynamic imports are run: https://github.com/thejameskyle/dynamic-import-reporter-loader

If what you say works, I could probably finalize the server component and be done in a day or two.

PR #27, opened 2017-04-02
Strawman: babel-plugin-transform-react-loadable

motiz88:
Hey there,

I made [babel-plugin-transform-react-loadable](https://github.com/motiz88/babel-plugin-transform-react-loadable) following [this random thought I had](https://twitter.com/motiz88/status/847215130429587457).

Weirdly enough, I don't actually have `react-loadable` set up in any of my projects at the moment. But I hadn't written a Babel plugin in a while, and the idea seemed valid enough, so I figured, what the heck. So this is me sharing it super early, but hopefully in a rather complete state, unless more rigorous testing shows otherwise.

I welcome any feedback, especially if anyone actually uses the plugin and has anything to report. The actual PR here is fairly minimal and should probably only be merged after it's rewritten and after we have evidence that this works in a real-world project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -234,6 +234,8 @@ directly because that would add it to the bundle in Webpack or Browserify.
 For `webpackRequireWeakId` it needs to be a function because
 `require.resolveWeak` does not exist in any tool other than Webpack.
 
+If you use Babel, check out [babel-plugin-transform-react-loadable](https://github.com/motiz88/babel-plugin-transform-react-loadable) for a more automated way of setting these properties wherever you use `Loadable()`.
+
 #### How do I avoid repetition?
 
 Specifying the same `LoadingComponent` or `delay` every time you use


thejameskyle:
Weird request, could we actually embed that transform in this repo? That way I can tell people to put this in their `.babelrc`:

```js
{
  "plugins": ["react-loadable/babel"]
}
```

And then it will be versioned with the project.

motiz88:
I would not mind that in the least.

faceyspacey:
@motiz88 absolutely love this babel plugin!

thejameskyle:
Added the plugin to the package. I started modifying it to make some changes and ended up changing most of it, but kept you as the committer @motiz88. Thank you for this.

aardito2:
Is it possible to allow the babel plugin to handle HOCs to wrap `Loadable`, such as the `MyLoadable` example in the readme?

The plugin currently breaks if it sees `Loadable` being called with an object expression that either does not contain a loader property, or encounters a property that does not have a key, such as object spread.

Using `Object.assign` instead of object spread circumvents the latter error. However, calling the HOC will not result in the addition of the server and webpack props to the returned `Loadable`.

thejameskyle:
heh... I was waiting for that question to come.

We'd need some way to identify which objects we should expand. 

For example, looking at just this code, how do we know we should match against the object with `loader`

```js
import MyReactLoadableWrapper from '...';
import MyComponent from '...';

const LoadableMyComponent = MyReactLoadableWrapper({
  loader: () => import('./foo'),
});
```

We could say "Match any object that has a `loader` property which is a function with an `import()` inside. But it would be _super_ confusing if that ever were to apply to something it wasn't supposed to.

So I was thinking about doing this instead:

```js
import {loader} from 'react-loadable';
import MyReactLoadableWrapper from '...';
import MyComponent from '...';

const LoadableMyComponent = MyReactLoadableWrapper({
  [loader]: () => import('./foo'),
});
```

Where `loader` is a Symbol exported from `react-loadable` which gets used everywhere and properly marks the object as being configuration supplied to react-loadable.

But I feel like that would be a controversial API decision... so idk

bwhitty:
> But I feel like that would be a controversial API decision... so idk

How many HOC will one make in a project? The use case is usually setting defaults and such, so I don't really see the use of a Symbol being that controversial. It would only be required when used within a HOC *and* when use the babel plugin. The existing API could be preserved for other use cases. It would definitely need some clear docs, but doesn't seem that complicated.

Regular component
```jsx

let LoadableMyComponent = Loadable({
  loader: () => import('./MyComponent'),
  LoadingComponent: MyLoadingComponent,
});
```

HOC
```jsx
import Loadable, {LoaderSymbol} from 'react-loadable';
import MyLoadingComponent from './MyLoadingComponent';

export default function MyLoadable(opts) {
  return Loadable({
    LoadingComponent: MyLoadingComponent,
    delay: 200,
    [LoaderSymbol]: opts.loader,
    ...opts
  });
}
```

An interesting side effect would be that spreading `opts` like that would duplicate the `loader` option in the options passed to `Loadable`, so it'd have to take precedence or possibly throw if strictness is preferred.

thejameskyle:
@bwhitty You misunderstood the API change I was suggesting. You'd have to use the symbol everywhere not just in the HOC

aardito2:
Wouldn't this just be a matter of doing `let loader = opts[loader] || opts.loader;` in the `Loadable` function? The babel plugin could stay as it is now for the `'loader'` string key when used in the arguments of the `Loadable` function, but would also check for the symbol globally.

PR #29, opened 2017-04-06
flushServerSideRenders, fix flow, hoist tryRequire

thejameskyle:


ctrlplusb:
Awesome James!  That was fast!

src/index.js
```diff
@@ -126,6 +128,10 @@ export default function Loadable<Props: {}, Err: Error>(opts: Options) {
     render() {
       let { pastDelay, error, Component } = this.state;
 
+      if (!isWebpack && serverSideRequirePath) {
```
smashercosmo: Hm, I guess this won't work if we serve bundled code on the server?

phyllisstein: I agree on this count. I was able to get this working with Webpack by removing the `!isWebpack` check and following a slightly modified version of the pattern @thejameskyle adumbrated [here](https://gist.github.com/thejameskyle/abecfe8ec2a7ce1e312a904527a31908) server-side (which you can see [here](https://gist.github.com/phyllisstein/a33964c7f393d7f846c3b11fe1659a76), if you're curious). After making those two changes, I have SSR working reliably.

smashercosmo: @phyllisstein would you be so kind to publish your full solution? Can't make it work, because flushServerSideRequirePaths always returns path to the main bundle, and not paths of the loaded chunks.

richardscarrott: @smashercosmo you could always wrap loadable and implement your own flushServerSideRequires, I did that [here](https://github.com/60frames/react-boilerplate/blob/master/src/components/lib/loadable/loadable.js) as I also bundle the server side code with webpack so needed the module ids rather than an absolute path.

thejameskyle: https://github.com/thejameskyle/react-loadable/pull/33

phyllisstein: @smashercosmo With @thejameskyle's wonderful work on #33 in place, the implementation gets much simpler. I dropped an MWE at <https://github.com/phyllisstein/react-loadable-mwe> anyway, in case it's at all helpful.

smashercosmo: Awesome) thx ya all)

PR #30, opened 2017-04-07
Fix for preload() example

dusnoki:
If we don't monitor if the component has been loaded then the component will load each time the button is hovered over.

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,14 +176,18 @@ let LoadableMyComponent = Loadable({
 });
 
 class Application extends React.Component {
-  state = { showComponent: false };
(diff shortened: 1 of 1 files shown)


thejameskyle:
No it won't, the internal logic of react-loadable prevents an import from loading multiple times https://github.com/thejameskyle/react-loadable/blob/master/src/index.js#L52

dusnoki:
Great then, I thought it would cause i left a console.log inside the onMouseOver function and it kept going over and over on hover.

thejameskyle:
I got u fam

PR #31, opened 2017-04-11
pass props to loader

djeeg:
Hi,

Thoughts on passing component props into the loader() method?

Ideally want to be be able to do something like this:

```
let Layout = Loadable({
    loader: (props) => {
        return Promise.all([
            System.import('./LayoutAsync'),
            System.import('./../logic/section1')
        ]).then(function([component, module]) {
            injectReducer(props.contextstore, {key: 'section1State', reducer: module.reducer});
            return component;
        })
    },
});
```

Thought could do stuff such as loading language specific resources

### DIFF
--- a/__tests__/Loadable.test.js
+++ b/__tests__/Loadable.test.js
@@ -120,3 +120,39 @@ test("resolveModule", async () => {
   await waitFor(200);
   expect(component.toJSON()).toMatchSnapshot(); // errored
 });
+
+let createPropsErrorLoader = delay => {
+  return async props => {
+    await waitFor(delay);
+    //console.log(props);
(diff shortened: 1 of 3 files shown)


thejameskyle:
Is there a reason you couldn't wrap `Loadable()` to do something like this:

```js
let MyLoadable = props => {
  return Loadable({
    loader: () => {
      return Promise.all([
        System.import('./LayoutAsync'),
        System.import('./../logic/section1')
      ]).then(function([component, module]) {
        injectReducer(props.contextstore, {key: 'section1State', reducer: module.reducer});
        return component;
      })
    },
  });
}
```

djeeg:
thanks for the tip

PR #32, opened 2017-04-14
postLoad hook

djeeg:
Thoughts on a post-load hook?

```
const Layout: React.SFC<any> = (props, context) => {
    let Component = Loadable({
        loader: () => System.import('./LayoutAsync'),
        serverSideRequirePath: (__dirname + '/LayoutAsync'),
        postLoad: (module) => {
            injectReducer(context.store, {key: module.stateKey, reducer: module.reducer});
            injectSagas(context.store, module.SagasList);
        },
    });
    return (<Component {...props} />);
}
```

(no tests as yet)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -19,7 +19,8 @@ type Options = {
   delay?: number,
   serverSideRequirePath?: string,
   webpackRequireWeakId?: () => number,
-  resolveModule?: (obj: Object) => LoadedComponent<Props>
+  resolveModule?: (obj: Object) => LoadedComponent<Props>,
+  postLoad: (obj: Object) => void
 };
 
 export default function Loadable<Props: {}, Err: Error>(opts: Options) {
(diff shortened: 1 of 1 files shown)


faceyspacey:
it would be nice to get the chunk ID in the arguments passed to the hook as well, so for instance you could load a css file corresponding to the chunk. I produced a fork of `extract-text-webpack-plugin` that generates css files for individual chunks that this could be used in conjunction with: 

https://github.com/faceyspacey/extract-css-chunk

Perhaps in this case, a `preLoad` hook would be best so they come in at closer to the same time.

faceyspacey:
I implemented your idea and gave you credit for it in the `React Universal Component` repo:

...though you always could have hacked `resolveModule` to do it ;)

thejameskyle:
I'm going to close this because I don't believe it solves the problem in a good way.

React Loadable components (like other higher order components) are meant to be declared statically at the top level:

```js
import React from 'react';
import Loading from './Loading';

const LoadableMyComponent = Loadable({
  loader: () => import('./MyComponent'),
  loading: Loading
});

export default class App extends React.Component {
  render() {
    return <MyLoadableComponent/>
  }
}
```

Inserting a callback function into something static like this is highly inconvenient if you care about the component tree at all. You would be forced into a situation where you have to update some kind of global state.

Instead, I recommend passing a `props.onLoaded` callback down through your Loadable component into either the imported component or the Loadable `render()` method a

```js
export default class MyComponent extends React.Component {
  constructor(props) {
    super(props);
    props.onLoaded();
  }
}
```

```js
const LoadableMyComponent = Loadable({
  loader: () => import('./MyComponent'),
  loading: Loading,
});

<LoadableMyComponent onLoaded={() => {...}}/>
```

PR #36, opened 2017-04-25
Add TypeScript typings.

odensc:
Mostly copy-pasted from the Flow types with modifications for TypeScript syntax.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,21 @@
+declare module "react-loadable" {
+	import * as React from "react";
+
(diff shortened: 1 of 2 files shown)


thejameskyle:
This is now out of date

odensc:
@thejameskyle Yeah, I've updated the ones on DefinitelyTyped instead. Thanks for the response anyway. :)

joshgeller:
@odensc Just a heads up that the test file in DefinitelyTyped doesn't compile. I'd take a shot at fixing it but I'm super new to TypeScript.

odensc:
@joshgeller Works for me... What's the error?

PR #37, opened 2017-04-27
add support for HMR, flushWebpackRequireWeakIds test, new flushRequires method

faceyspacey:
This commit is a long time coming. This repo has the comments turned off, so I'm going to take a second to describe the journey to this point:

Originally my commit here had the code for comprehensive chunk flushing support, but then I realized it made the most sense to follow the Unix philosophy and create another package since the API boundaries were so clean. 

As a result, the following packages handle javascript and css chunk flushing by being passed the requires of `react-loadable`:

https://github.com/faceyspacey/webpack-flush-chunks
https://github.com/faceyspacey/extract-css-chunks-webpack-plugin

And here's a new boilerplate putting all 3 of these packages together:
https://github.com/faceyspacey/flush-chunks-boilerplate
*it works now, even without this PR. Give it a try!*

A LOT of work has gone into making those packages and the boilerplate feature-complete, tested and well-documented. This commit would have come several weeks ago, except handling CSS *(the creation of multiple css chunk files)* and support for css HMR ended up being quite the unanticipated trek. But I think the result is actually a solution that gives competing CSS solutions a run for their money. My notes on the matter are in the `extract-css-chunks-webpack-plugin` [readme](https://github.com/faceyspacey/extract-css-chunks-webpack-plugin).

In general, every which scenario I could think of has been covered in `webpack-flush-chunks` and its corresponding boilerplate: babel building, universal webpack bundling, development, production, embedding your chunks as strings or React Components, embedding CSS instead of stylesheets, HMR, sending the smallest amount of CSS over the wire and more. As I said, I originally intended for the `webpack-flush-chunks` package to be part of React Loadable, but after you added the 2 flushing functions it just made the most sense to keep it separate, so that other packages can rely on it as well.

If you have the time, I'd love your feedback on the above tools. My hope is for them to become official companion tools to *React Loadable* for their given purposes.

...What was left after I removed the chunk flushing are the following things in this commit:

- support for HMR by storing the updated component in the closure rather than instance `state`
- the missing test for `flushWebpackRequireWeakIds()`
- a proposed `flushRequires` method which provides a clean API to perform module flushing regardless of the environment. It helps make documentation for using `webpack-flush-chunks` a bit easier to digest.
- and a `wallaby.js` file because that's central to my workflow and I'd like to provide any help you need in maintaining this package 💯

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,8 +11,11 @@ A higher order component for loading components with promises.
 - Supports server-side rendering via a dynamic `require()`
 - Eagerly preload components when needed
 - Support for requiring synchronously from webpack when available
+- Supports chunk flushing via `webpack-flush-chunks` for complete server-side rendering
 
 Example Project: https://github.com/thejameskyle/react-loadable-example
+
+Example Chunk Flushing Project: https://github.com/faceyspacey/flush-chunks-boilerplate
 
 Introductory blog post: https://medium.com/@thejameskyle/react-loadable-2674c59de178#.6h46yjgwr
 
@@ -199,18 +202,20 @@ class Application extends React.Component {
 }
 ```
 
-#### `flushServerSideRequirePaths` / `flushwebpackRequireWeakIds`
+#### `flushServerSideRequirePaths` / `flushwebpackRequireWeakIds` / `flushRequires`
 
 In case you are rendering server-side and want to find out after a render cycle
 which `serverSideRequirePath`'s and `webpackRequireWeakId`'s were actually
 rendered, you can use `flushServerSideRequirePaths` or
-`flushWebpackRequireWeakIds` to get an array of them.
+`flushWebpackRequireWeakIds` to get an array of them. Or use `flushRequires` and *React Loadable*
+will decide for you what should be flushed.
 
 ```js
 import ReactDOMServer from 'react-dom/server';
 import {
   flushServerSideRequirePaths,
-  flushWebpackRequireWeakIds
+  flushWebpackRequireWeakIds,
+  flushRequires
 } from 'react-loadable';
 
 let app = ReactDOMServer.renderToString(<App/>);
@@ -218,9 +223,56 @@ let serverSideRequirePaths = flushServerSideRequirePaths();
 // ["/path/to/component.js", "/path/to/other/component.js"]
 let webpackRequireWeakIds = flushWebpackRequireWeakIds();
 // [1, 2]
+let weakIdsOrRequirePaths = flushRequires();
+// ["/path/to/component.js", "/path/to/other/component.js"]
+// or
+// [1, 2]
+// depending on the environment (Webpack vs. Babel)
 ```
 
 > **Note:** These are flushed individually, one does not affect the other.
+
+## Complete SSR Handling /w `webpack-flush-chunks`
+
+If you want to bypass having to determine which chunks are needed to synchronously render 
+your modules/components on the client, we highly recommend using our companion package: 
+`webpack-flush-chunks`.
+
+```js
+import path from 'path';
+import React from 'react';
+import ReactDOMServer from 'react-dom/server';
+import ReactLoadable from 'react-loadable';
+import flushChunks from 'webpack-flush-chunks';
+import App from '../src/components/App';
+
+export default function serverRender(webpackStats) {
+  return (req, res, next) => {
+    const app = ReactDOMServer.renderToString(<App />);
+    const moduleIds = ReactLoadable.flushRequires();
+
+    const { Js, Styles } = flushChunks(moduleIds, webpackStats, {
+      before: ['bootstrap', 'vendor'],
+      after: ['main'],
+      rootDir: path.resolve(__dirname, '..')
+    });
+
+    const html = ReactDOM.renderToStaticMarkup(
+      <html>
+        <body>
+          <Styles />
+          <div id="root" dangerouslySetInnerHTML={{ __html: app }} />
+          <Js />
+        </body>
+      </html>
+    );
+
+    res.send(`<!DOCTYPE html>${html}`);
+  }
+}
+```
+
+Checkout [faceyspacey/webpack-flush-chunks](https://github.com/faceyspacey/webpack-flush-chunks) to learn more.
 
 ## Babel Plugin
 
--- a/README.md
+++ b/README.md
@@ -227,7 +227,7 @@ let weakIdsOrRequirePaths = flushRequires();
 // ["/path/to/component.js", "/path/to/other/component.js"]
 // or
 // [1, 2]
-// depending on the environment (Webpack vs. Babel)
+// depending on the environment (Babel vs. Webpack)
 ```
 
 > **Note:** These are flushed individually, one does not affect the other.
--- a/__fixtures__/stats.js
+++ b/__fixtures__/stats.js
@@ -1,0 +1,46 @@
+export const stats = {
+  assetsByChunkName: {
+    bootstrap: ["bootstrap.js", "bootstrap.no_css.js"],
+    vendor: ["vendor.js", "vendor.no_css.js"],
+    main: ["main.js", "main.no_css.js", "main.css"]
+  },
+  chunks: [
+    {
+      id: 0,
+      files: ["0.js", "0.no_css.js", "0.css"]
+    },
+    {
+      id: 1,
+      files: ["1.js", "1.no_css.js", "1.css"]
+    }
+    // chunk with id: 2 intentionally missing to test against invalid stats
+  ],
+  modules: [
+    {
+      id: "qwer",
+      name: "./src/Components/Example.js",
+      chunks: [0]
+    },
+    {
+      id: "asdf",
+      name: "./src/Components/Foo.js",
+      chunks: [1]
+    },
+    {
+      id: "zxcv",
+      name: "./src/Components/Bar.js",
+      chunks: [2]
+    }
+  ],
+  publicPath: "/static/"
+};
+
+export const babelFilePaths = [
+  "./src/Components/Example.js",
+  "./src/Components/Foo.js",
+  "./src/Components/Bar.js"
+];
+
+export const webpackModuleIds = ["qwer", "asdf", "zxcv"];
+
+export const rootDir = "/Users/jamesgillmore/App";
--- a/__tests__/Loadable.test.js
+++ b/__tests__/Loadable.test.js
@@ -3,7 +3,14 @@ declare var test: any;
 import path from "path";
 import React from "react";
 import renderer from "react-test-renderer";
-import Loadable, { flushServerSideRequirePaths } from "../src";
+import Loadable, {
+  flushServerSideRequirePaths,
+  flushWebpackRequireWeakIds,
+  flushRequires
+} from "../src";
+
+// normalize the required path so tests pass in all environments
+let normalizePath = path => path.split("__fixtures__")[1];
 
 let waitFor = (delay: number) => {
   return new Promise(resolve => {
@@ -121,7 +128,7 @@ test("resolveModule", async () => {
   expect(component.toJSON()).toMatchSnapshot(); // errored
 });
 
-test("server side rendering flushing", async () => {
+test("babel: server side rendering flushing", async () => {
   let createLoadable = name => {
     return Loadable({
       loader: createLoader(400, null, new Error("test error")),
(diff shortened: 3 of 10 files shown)


thejameskyle:
This is great work, thank you for taking it on.

I had been thinking about a similar utility built into react-loadable itself in a separate `react-loadable/server` library.

```js
import App from './app';
import {renderToString} from 'react-dom/server';
import {flushWebpackRequireWeakIds} from 'react-loadable';
import {getLoadedWebpackChunks} from 'react-loadable/server';
```

It would do a lot less than you've done here, it would only extract loaded chunks from webpack stats.

```js
let stats = require('./webpack-stats.json');
let app = renderToString(<App />)
let webpackRequireWeakIds = flushWebpackRequireWeakIds()
let loadedChunks = getLoadedWebpackChunks(webpackRequireWeakIds, webpackStats);
```

You would then render these on your own:

```js
let styles = loadedChunks.filter(chunk => chunk.fileName.endsWith('.css'));
let scripts = loadedChunks.filter(chunk => chunk.fileName.endsWith('.js'));

res.send(`
  <!doctype html>
  <html>
    <head>
      <title>Hello World</title>
      ${styles.map(chunk =>
        `<link rel="stylesheet" href="/assets/${chunk.fileName}"></script>`
      ).join('\n')}
    </head>
    <body>
      <div id="root">${app}</div>
      ${scripts.map(chunk =>
        `<script src="/assets/${chunk.fileName}"></script>`
      ).join('\n')}
    </body>
  </html>
`)
```

This gives people the minimal api that is maximally flexible. It's also really easy to teach people.

What do you think?

faceyspacey:
Glad u like it.

This actually existed right at 'react-loadable/server' until a day ago when it became time to ship and final decisions had to be made. The idea was for people familiar with using ReactDOM on the server to feel right at home.

As far as the minimal API, I'm not sure if you saw but you can also get an array of style sheets and an array of js scripts.

I like the idea of manually being able to get access to the chunks so that people have the opportunity to get any custom assets they may have put in there. Who knows how many ways people have found to use webpack that I'm not privy to. Though based on my work with extract CSS chunks, my hunch is not many packages exist that add additional assets per dynamic chunk. If so we should catalog them as they make sense in the RL family of companion packages. Full on code-splitting is an underrepresented opportunity with webpack. It's there but until ur post came out--particularly the module flushing concept--most people didn't know what to do about it to really get it right. (In my previous apps I just gave up on the ssr aspect and saved bytes where I could via additional async requests). I think we are finally picking up steam in terms of idiomatic best practices we can present to the community.

How about this: we run for 2 weeks without this capability in RL and I'm going to do a little promotion of my package (maybe you can help me with that). And then--if what ur saying is u'd like some minimal API within RL itself--I'll hook RL up and add the minimal API u mentioned. It won't take me long.

As far as "teaching people," it's really all about the webpack configs. The application code is easy peasy with either of our APIs, but the webpack config--although it's one time--is surgical. I anticipate lots of issues from people missing one configuration line. Configuring CSS is likely going to be the biggest pain point. That said for the very meticulous, it all should be a breeze. For the lazy and beginners (and the smart ones), it's just a question of starting out with the boilerplate and never leaving it.

So I'm not sure if u want duplicate info for all that configuration in ur repo.

...It all depends on what ur goals are. Are ur goals to blow ur line of tools up, or are you too busy (again, I noticed comments off on this repo, I'm just assuming U got a lot on ur plate). I am and have been interested in promoting tools that offer what Next.js does but without the "framework." We need to amp up the momentum on building raw npm React apps before we have another "Meteor" on our hands. In my view, Next.js is a problem because it brings new functionality we should be able to achieve on our own without its limitations. Specifically: server-rendered code-splitting with prefetching. That's its self-described main draw. I have prefetching tools for routers in conjunction with the RL family coming out next. Next.js isn't the only one that gets to play :).

So another path I pondered--and keep in mind nothing is sealed yet since very few know about these packages--was calling it 'react-loadable-flush-chunks'. I mean I really wanted to insure what I worked on had as much potential to be used as possible, to be the defacto solution for its market.

With all that in mind, the decisions to be made rest on how U would like things to pan out. Perhaps u don't care much to order the messy world of npm packages. Or perhaps u do. The way I see it is it takes developers giving a shit to find order for everyone in all the confusion. That's done through package naming, shared path segments if you will: webpack-dev-middleware vs webpack-hot-middleware, etc. It's done with clear API contracts which our packages clearly have. It's a messy dog eats dog world out here lol, somebody needs to make sense of it so we don't all repeat the same footsteps over and over.

In short, with ur promotion capabilities we could takedown this segment of the market. It all depends on whether u have the motivation to. I can signup to do responsive issue handling. I can and want to do a lot of maintenance. I'm between contract projects and plan to devote myself to open source for a long while as it's something I haven't had the chance to do much of. I also have a long vision for packages I'm building--basically extracting lots of stuff from paid projects. I'm actually going to try to make a career out of (learning videos etc) it so I don't have to go back to contract work ;). In short, if u don't have time to deal with it, I can handle ur repo issues and will have availability for a long while. 

Do you wanna expand the RL family together? If so, let me know what package names and locations for the code you think is best that can get the most maximum impact. Perhaps we do just wanna put it all back into RL proper. I'm open to it and whatever path will get the most traction.

faceyspacey:
Also, if I didn't make my perspective fully clear: basically it's a new way of doing things. The motto is "code splitting everywhere." With that, things like HMR, CSS, routing and likely a lot more become important. Similar to using Next.js, this *way of doing things* needs to be seamless and well-documented. Having one single package goes a long way to simplifying things for people. I think we can do the same with multiple packages (the unix/npm philosophy is just better that way), but consistent naming.

thejameskyle:
Sorry, I'm visiting family right now so I'm not very responsive at the moment.

---

The new model for code splitting is this:

1. Split in as many locations as possible
2. Use tooling to automate bundle optimization

That automation can come in one of two forms:

1. Server-side rendering to automatically send the right bundles down
2. Report metrics from the client to collect data about when bundles are needed (does not require SSR)

Either way we're going to need code running on the server controlling the bundles sent to the client.

That code running on the server will be heavily tied to the code used on the client and the way that the code is compiled with Babel/Webpack.

Since the client-side code, compiler code, and server-side code is tightly coupled, they should live together within a single module. Splitting them apart into separate modules creates the appearance of being decoupled without actually being decoupled. These tools aren't very useful on their own.

---

As for the api, I did read all of your code and I know what you're going for. But I'm afraid you're coding yourself into a corner api-wise by taking on turning chunks into link/script tags. The api isn't very flexible, and people will have to ask you for new features instead of coding it themselves.

By forcing people to do this:

```js
let styles = loadedChunks.filter(chunk => chunk.fileName.endsWith('.css'));
let scripts = loadedChunks.filter(chunk => chunk.fileName.endsWith('.js'));

styles.map(chunk =>
  `<link rel="stylesheet" href="/assets/${chunk.fileName}"></script>`
).join('\n')

scripts.map(chunk =>
  `<script src="/assets/${chunk.fileName}"></script>`
).join('\n')
```

You actually make the api much more flexible, which removes the need for people to ask for features.

faceyspacey:
I added what you mentioned last week:

https://github.com/faceyspacey/webpack-flush-chunks#low-level-api-flushfiles

https://github.com/faceyspacey/webpack-flush-chunks/commit/dc98ca49e2ba2ea47c2dda276bbae8fee18dca8e#diff-339736190ca5b7784737038adcb0085b


It pretty much was just the case of exposing a function we already have. It's not exactly what you wrote in that it doesn't give you direct access to the `chunk` objects, but rather just the filenames. If that's absolutely needed we can go that route as well. I chose this route because it maintained implementation the succinctness and concept I was going for.

Anyway, I was initially against complicating the API with 2 api levels, but figured it would be good to look at during this stage, and because if it was phrased in the documentation as an *after-thought* it wouldn't lead too many astray. That said, you also seem to want it as the primary thought, if not only thought. 

The obvious conclusion is we move the new low level API into RL, and the remaining can be a separate package--cuz hey it's a free world. 

----
Regarding:
> Since the client-side code, compiler code, and server-side code is tightly coupled, they should live together within a single module. Splitting them apart into separate modules creates the appearance of being decoupled without actually being decoupled. These tools aren't very useful on their own.

That basically assumes there won't be many other packages like RL. I think there will, if there aren't already. Of course there will. I have many features you probably won't merge right now that might just have to become their own package. I had something very close before RL came out. So did lots of people. Mine was able to toggle to synchronous loading when it could.  This is just the most heavily promoted. So I disagree that chunk flushing as a standalone feature is not important. It is. All `webpack-flush-chunks` currently needs is an array of module IDs. Any code splitting package can produce that.

That said, it's all about timing. There's no need to stoke that fire instead of RL's fire. When the time comes, things can be broken out for other packages to use. ...That's a slightly more nuanced perspective on the whole thing. ...and other packages can include RL and use its chunk flushing tools without its component if it really wants--so I'm not too worried.

My main thing is I wanna get this out. I want it in circulation. I'm down to pump up the RL repo itself by merging the code into this. What do you need to push this forward?  I mean I think I know. If I don't hear from you, I'm gonna make a PR directly to this repo with the API you want (filenames instead of chunks until you chime in otherwise). I'm gonna move my automation to another package, and people can use that if they want the utmost in developer experience or automagic--however u wanna see it. And you get your dream of that professional right level (not too high, not too low) of abstraction. *A set of primitives, even the most sophisticated of React team members would cosign.* I gotchu brother. Coming right up.

faceyspacey:
What i described has been pushed to this PR with tests. 

Perhaps it wasn't really an issue to you, but here's why I went for an array of file name strings instead of chunk objects:

https://www.dropbox.com/s/j74q959hpdq7yct/Screenshot%202017-05-02%2006.24.17.png?dl=0

That's a screenshot of what those objects look like. Here are the issues with it:
- each chunk object only has an id, such as `0`; only named entries have a name, as seen in the `names` array (which I guess can be multiple)
- it's the `chunk.files` array which has the files, of which there will be multiple file types per chunk
- there's not much other useful info in each `chunk` object. The `hash` is for the javascript file, which will be different than the `[contenthash]` used for css files. So it's not like there's one consistent hash--again, it's the filenames that matter.
- all dynamic chunks will have `entry` and `initial` set to `false`, which makes those keys useless
- I expanded `stats.chunks[0].origins` so you could see what's inside. It's the parent file that initiated the split. This info plus the `parents` key which has the ID of the parent chunk might be useful to advanced users, *but my thinking is to let it go*. 

So there you go. You decide if it's worth producing chunk objects instead of filenames and I'll make it happen. 

Otherwise what you got now is an API like this:

```js
import ReactDOM from 'react-dom/server'
import ReactLoadable from 'react-loadable/server'

const app = ReactDOM.renderToString(<App />)
const moduleIds = ReactLoadable.flushRequires()
const files = ReactLoadable.flushFiles(moduleIds, stats)

let styles = files.filter(file => file.endsWith('.css'))
let scripts = files.filter(file => file.endsWith('.js'))

res.send(`
  <!doctype html>
  <html>
    <head>
      <title>Hello World</title>
      ${styles.map(name =>
        `<link rel="stylesheet" href="/assets/${name}"></script>`
      ).join('\n')}
    </head>
    <body>
      <div id="root">${app}</div>

      <script src="/assets/bootstrap.js"></script>

      ${scripts.map(name =>
        `<script src="/assets/${name}"></script>`
      ).join('\n')}

      <script src="/assets/main.js"></script>
    </body>
  </html>
`)
```

Of course, it's now up to the user to use the `bootstrap` technique or do the `window.render` technique as facilitated in a previous PR by @richardscarrott . And that's in addition to getting the proper hashes for these files in production. But that's the whole idea here--a base primitive that offers supreme flexibility and does very little, but does it well while making its main utility clear.

faceyspacey:
@thejameskyle not sure if you saw this. Let me know what else you need. Documentation? Boilerplate update? When you approve of it, I can work on those 2 things.

panjiesw:
I came across HMR issue with dynamically loaded components with `react-loadable` (without SSR) and found this. I've tried this pull request by @faceyspacey locally and the HMR is now working with those components. Hope this get merged and released eventually.

faceyspacey:
@thejameskyle hello?? this has been almost 3 weeks now. We coulda been the first to release this, but now Next.js 3 basically has. Developers that aren't ok with using a framework like Nextjs need this. Are we doin this or not brotha?

faceyspacey:
Also if you wanna do this, we should discuss perhaps coinciding this with a 2.0 that changes the api/props to simpler more memorable names like we're seeing in other similar components that have been released recently (e.g. Vue's rendition, Nextjs 3.0's "dynamic" components, etc). Similarly I think just a `flushRequires` method should replace the 2 long webpack/babel-specific ones.

I'm more than willing to do all the work to make this take up as little of your time as possible (documentation, boilerplates). Both basically are already done, we just need to discuss where they go.

faceyspacey:
Also, in interest in really getting this right--and I really don't care with the code goes--you may wanna reconsider whether React Loadable houses the server-side chunk flushing. Here's another very interesting package (which I first learned about in the above link to Nextjs' "dynamic" component): https://github.com/xtuc/async-reactor

It doesn't buffer the module IDs, but I'll probably provide a PR that does. In short, it's another take on imported components. In general, it's looking like there is need for a non-component based way to make async imports while recording the module IDs. 

So with that said, I totally agree that the lower-level API as you described is the way to go (webpack-flush-chunks now has that). But I don't agree that chunk flushing *needs* to live in whatever package provides async importing functionality. I'm fine with it, but in the long-run it's my opinion that everyone would benefit from a separate package that just handles that aspect. 

The main thing is we have something the industry needs. Being that this is a more general solution than having to use Next.js, this is the most significant release of this feature. Since you sparked this "revolution" if you will, the most important part is the impact of your *voice* to present this to javascript developers (e.g. Medium post, etc). See, even Arunoda mentioned you in his release of Nextjs 3.0's dynamic async imports yesterday: https://zeit.co/blog/next3-preview#dynamic-imports

In interest of time, I'll let you know what my interest is here: I have several other packages I'd like to piggy back on this release. Otherwise I wouldn't be waiting around this long. I need to get them out. So let's make magic happen my friend. These are key introductions to the React world.

Bamieh:
i'd write my whole life story in less paragraphs!

ctrlplusb:
That isn't a very constructive addition to this conversation @Bamieh 🙁

@faceyspacey is clearly giving a lot of his time to this issue, so please respect that by avoiding sarcasm. If it's too much to onboard rather ask for a tldr summary of above. I am sure they would be happy to oblige.

Bamieh:
@ctrlplusb its a joke, i am interested in the solution so i had to read the whole thing, its still too much writing non the less.

lourd:
I'd like to second many of the points that @faceyspacey made, especially about this library being the main alternative to Next.js for code-split, server-rendered React apps, and the importance and opportunity to offering the community an alternative to a full-stack framework solution.

It's clear @faceyspacey is amped about this problem space and making progress, and there's a clear need for this library & a couple other building blocks within the community. @thejameskyle, what are you thinking? How you doing? How's your bandwidth look for the summer?

Total aside: what do y'all think about how the Vue loader and `next/dynamic` libraries take the actual call to `import` an option (`component: import('./MyComponent') `), instead of a function calling import, like this library does? It looks like their implementations create the promise, but don't resolve it until the component is actually rendered. What are y'alls opinions on the tradeoffs?

thejameskyle:
I focus on the open source projects that I am paid to focus on. I'm in the middle of a 7,500 mile move across the Pacific Ocean right now so I don't have much bandwidth otherwise.

If someone could simplify this down for me, I can make a decision about it. I just don't have time to work this all out on my own right now.

Also please be civil with one another. I don't tolerate people being jerks on my own projects, I just block without warning.

lourd:
Thank you for your prompt reply!

Since your time in the near & mid-term is going to be occupied with more important personal commitments, and assuming this project falls into the non-paid category of your OSS projects, would you be willing to add maintainers from the community? I think the library has clear room to grow and would benefit from some active development, especially within these next couple months. No worries if you're not down, folks can fork or turn to other libraries or whatever, just looking for clarity.

One step in active development would probably be re-enabling issues. As much as a time suck as they can be, it's essential for folks to be able to ask for help, and discussions to be had outside of PRs.

@faceyspacey, I think we'd all benefit from a succinct summary of your proposal at this point 🙂 Thanks in advance, we know you've written a lot already.

smashercosmo:
I want to add my 5c to this discussion. @faceyspacey has done an awesome work in this PR and in his related packages, but I would assume, that it would be easier for @thejameskyle to review and merge these additions if they are split into separate smaller PRs.

faceyspacey:
Hey guys, given the @thejameskyle 's time constraints I decided a while ago to go do my own thing. Here are the fruits:

I wanna thank you James for drawing my attention to this opportunity. If you like what I've come up with, a tweet would be cool. 

Peace out fellas

PR #38, opened 2017-05-04
add comment for webpack async chunks

restrry:


restrry:
@thejameskyle also I faced a problem that babel plugin doesn't process `require.ensure`, so I had to add that functionality in my fork and use it in my project. do you want me send pr or you think this functionality is redundant?

thejameskyle:

PR #39, opened 2017-05-13
Add webpackChunkName option and update babel plugin

simenbrekken:
This builds on ideas from `flushServerSideRequirePaths` but focuses on chunks rather than file-system paths. 

By utilizing [`webpackChunkName`](https://webpack.js.org/guides/code-splitting-async/#chunk-names) alongside a new method `flushWebpackChunkNames`, you'll be able to generate list of chunks that were included in a server-side render. This can then be used with something like [`webpack-manifest-plugin`](https://github.com/danethurber/webpack-manifest-plugin) to map chunk names to chunk output filenames.

**LoadingComponent.js**

```js
const Page = withLoadable({
  loader: () => import('./MyComponent' /* webpackChunkName: "my-component" */),
});
```

**server.js**

```js
import { flushWebpackChunkNames } from 'react-loadable';

...
const manifest = require('./build/manifest.json');
const chunks = flushWebpackChunkNames().map(name => `/${name}.js`);
const scripts = chunks.map(path => manifest[path]);

renderToStaticMarkup(
  <html>
    <head>
      {scripts.map(src => <script src={src} />)}
    </head>
  </html>
);
```

It's possible to manually specify the chunk name using the `webpackChunkName` option as with `serverSideRequirePath`. The updated babel plugin includes generation of this property based on the comment contents.

I haven't updated the `README` yet incase you decide against this new feature or want to change it but I'd be more than happy update that as well.

### DIFF
--- a/__tests__/Loadable.test.js
+++ b/__tests__/Loadable.test.js
@@ -3,7 +3,10 @@ declare var test: any;
 import path from "path";
 import React from "react";
 import renderer from "react-test-renderer";
-import Loadable, { flushServerSideRequirePaths } from "../src";
+import Loadable, {
+  flushServerSideRequirePaths,
+  flushWebpackChunkNames
+} from "../src";
 
 let waitFor = (delay: number) => {
   return new Promise(resolve => {
--- a/__tests__/__snapshots__/babel-plugin.test.js.snap
+++ b/__tests__/__snapshots__/babel-plugin.test.js.snap
@@ -37,6 +37,26 @@ let LoadableMyComponent = Loadable({
 });"
 `;
 
+exports[`add-webpack-chunk-name-property 1`] = `
+"import Loadable from \\"react-loadable\\";
+let LoadableMyComponent = Loadable({
+  loader: () => import(\\"./MyComponent\\" /* webpackChunkName: \\"my-component\\" */),
+  webpackChunkName: \\"my-component\\",
+  webpackRequireWeakId: () => require.resolveWeak(\\"./MyComponent\\"),
+  LoadingComponent: MyLoadingComponent
+});"
+`;
+
+exports[`add-webpack-chunk-name-property 2`] = `
+"import Loadable from \\"react-loadable\\";
+let LoadableMyComponent = Loadable({
+  loader: () => import( /*webpackChunkName: \\"my-component\\"*/\\"./MyComponent\\"),
+  webpackChunkName: \\"my-component\\",
+  webpackRequireWeakId: () => require.resolveWeak(\\"./MyComponent\\"),
+  LoadingComponent: MyLoadingComponent
+});"
+`;
+
 exports[`not-overwrite-existing-properties 1`] = `
 "import Loadable from \\"react-loadable\\";
 let LoadableMyComponent = Loadable({
--- a/__tests__/babel-plugin.test.js
+++ b/__tests__/babel-plugin.test.js
@@ -122,3 +122,37 @@ test("track-imports-correctly", () => {
     )
   ).toMatchSnapshot();
 });
+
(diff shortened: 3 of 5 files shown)


Bamieh:
but why would i want to map the chunk names to the paths and include them in the head? since `flushServerSideRequirePaths` is already the mapped filename why not just use it?

simenbrekken:
@Bamieh Sorry for the late reply. Because there's no direct correlation between source file and output chunks.

```js
const Foo = () => import(/* webpackChunkName: "/foo" */ './Foo')
const Bar = () => import(/* webpackChunkName: "/bar" */ './Bar')
const Baz = () => import(/* webpackChunkName: "/bar" */ './Baz')
```

Would output:

- `foo.js` (containing the Foo module)
- `bar.js` (containing Bar and Baz modules)

faceyspacey:
Awesome Idea! ...I provided a PR that achieves the same while not requiring you to name your chunks:

https://github.com/thejameskyle/react-loadable/pull/37

https://github.com/faceyspacey/webpack-flush-chunks

I jump through a lot of hoops to automate this for the user. But what you got is very compelling. Got me thinking...

faceyspacey:
How do you name dynamic chunks though? Is there a webpack configuration to name them so they aren't named `0.js`, `1.js`, etc? I don't think there is. It's also not possible in without a mechanism to name the dynamic import within code, as the id numbers can easily change.

If you read Webpack's creator's comment here, the new `import` spec doesn't allow for specifying a chunk name:

https://github.com/webpack/webpack/issues/1949#issuecomment-177108846

-----

However, the old legacy way allows you to:

https://webpack.js.org/guides/code-splitting-async/#require-ensure-

```js
require.ensure([], (require) => {
    const MyComponent = require('./MyComponent')
}, 'chunkName')
```

...However, to work with React Loadable, `MyComponent` in the example needs to be the resolution of a promise. So you need to do this:


```js
const Example = ReactLoadable({
  webpackRequireWeakId: () => require.resolveWeak('./MyComponent'),
  loader: () => { 
     new Promise((resolve, reject) =>
        require.ensure(
           [],
           (require) =>  resolve(require('./MyComponent')),
           (error) => reject(error), 
           'chunkName'
        )
     )
  },
```

So the obvious downsides of that are that you have to use the legacy `require.ensure` technique which the maintainers of Webpack are moving away from and have deprecated. That's unfortunate for this very use case. In addition, it's extremely verbose.

simenbrekken:
@faceyspacey If you read my description you'll see that I'm referring to https://webpack.js.org/guides/code-splitting-async/#chunk-names which does exactly this.

faceyspacey:
@simenbrekken crazy, I was wondering what those comments were about.

faceyspacey:
@simenbrekken your idea is implemented and I gave you credit for it in the `React Universal Component` repo:

simenbrekken:
@faceyspacey Thanks! Definitely going to checkout [react-universal-component](https://github.com/faceyspacey/react-universal-component)

aniruddhashevle:
@faceyspacey : Thanks for the solution. I'm able to build the chunks using your soln. But I'm facing some issues while resolving component's path. Please refer below.

I've defined const yoPath has the link for the resp component.

`const yoPath = 'src/components/scenes/dashboard/dashboard';`

screen shot 2017-08-22 at 3 17 49 pm

In the console,

screen shot 2017-08-22 at 3 19 15 pm

and the chuckfile,
screen shot 2017-08-22 at 3 19 58 pm

Please let me know if I'm doing anything wrong.

refanbanzhang:
> ```js
> /* webpackChunkName: "/foo" */ './Foo'
> ```

Nice work!

PR #41, opened 2017-05-20
WIP: Webpack server / fix for checksum issue

ndreckshage:
Awesome project!

PR not meant to be merged as is. Wanted to start a discussion re: webpack server & checksum mismatch

- I use webpack on server. Currently the `isWebpack` check returns false in webpack server env.
- There is a general issue with react checksum not matching on master. Server renders, but then client requests the JS, and empties dom.

This lets lets us flush our chunk names, so that we can load all the scripts we need before rendering client side.

```
const Component = Loadable({
  loader: () => import(/* webpackChunkName: "myComponent" */'myComponent'),
  webpackRequireWeakId: () => require.resolveWeak('myComponent'),
  chunkName: 'myComponent',
});
```

can play with it in my starter kit https://github.com/HumbleSpark/sambell - scoped this fork to `@humblespark/react-loadable`

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,10 @@
+## FORK of react-loadable
+
+Key differences:
+- Supports running webpack on server code.
+- Adds a flush for webpack chunk names.
+- Default loading component.
+
 # `react-loadable`
 
 A higher order component for loading components with promises.
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
-  "name": "react-loadable",
-  "version": "3.3.1",
+  "name": "@humblespark/react-loadable",
+  "version": "3.4.1",
   "description": "A higher order component for loading components with promises",
   "main": "lib/index.js",
   "author": "James Kyle <me@thejameskyle.com>",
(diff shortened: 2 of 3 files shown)


ndreckshage:
didn't see this (https://github.com/thejameskyle/react-loadable/pull/39) before doing this.

webpack server stuff still relevant though. i think we should be checking on standard is server checks `typeof window` etc, rather than relying on iswebpack

faceyspacey:
you're issue with `isWebpack` returning false is likely because `react-loadable` is added to externals. If you are removing node_modules from externals, you need to skip `react-loadable` so it still runs within Webpack's context. Like this:

```js
const externals = fs
  .readdirSync(modeModules)
  .filter(x => !/\.bin|react-loadable/.test(x))
  .reduce((externals, mod) => {
    externals[mod] = `commonjs ${mod}`
    return externals
  }, {})
```

ndreckshage:
oh awesome you are probably right. will try that later. thanks!

LeoIannacone:
Hey guys, any news about this PR, I'm having checksum mismatch and I'd like to use your fork @ndreckshage (I wasn't able to install via npm btw)

faceyspacey:
This should solve it for you: https://medium.com/@faceyspacey/code-cracked-for-code-splitting-ssr-in-reactlandia-react-loadable-webpack-flush-chunks-and-1a6b0112a8b8

...just happened to have wrote that a few seconds ago.

ndreckshage:
Thanks @faceyspacey - I'll take a look at that later. Closing this PR though, seems like there are better solutions.

PR #42, opened 2017-05-25
Pass props destined for the component that is being loaded into `LoadingComponent` for interim display purposes

alechill:
Passing the `LoadableComponent`'s props into the `LoadingComponent` rather than just `isLoading/error/pastDelay` allows the `LoadingComponent` to be much smarter about what it can display.  

Rather than simply acting as a standard message or indicator (or having to create multiple LoadingComponents tailored for different scenarios), it could act as a generic placeholder, while still being able to show customised messaging, and take on the same dimensional or visual style properties etc albeit in a lightweight and generic interim component. 

This can aid the loading to feel much more seamless. 

As the `LoadingComponent` is passed by class reference to `Loadable` rather than a component instance, I cant think of any way to wrap the Loadable into another HOC that can affect such props within the LoadingComponent itself, based on those props that are destined for the final component?

For example, I cant see how another HOC could get the expansion props destined for the `MapCardContent` component to be usable by `CardContentLoading` component in the interim (The following was tested and works with the proposed change by the way)

```TypeScript
import * as React from 'react'
import Loadable, {LoadableComponent, LoadingComponentProps, LoadedComponent} from 'react-loadable'
import {CardContentExpander, CardContentExpanderProps} from '../../CardContentExpander/CardContentExpander'
import * as lazy from './MapCardContent'

export const CardContentLoading = (props: CardContentExpanderProps & LoadingComponentProps) =>
  <CardContentExpander expanded={props.expanded} gridSpanX={props.gridSpanX} gridSpanY={props.gridSpanY}>
    {props.isLoading && <span>Loading...</span>}
    {props.error && <span>ERROR</span>}
  </CardContentExpander>

type LazyProps = lazy.MapCardContentProps // (extends CardContentExpanderProps)
type LazyComponent = typeof lazy.MapCardContent

export default Loadable<LazyProps, LazyComponent>({
  loader: (): Promise<LazyComponent> => System.import<LazyComponent>(/* webpackChunkName: "map" */ './MapCardContent'),
  LoadingComponent: CardContentLoading
})
```

Also fixed issue with fixture paths in snapshots, as were failing if you were on a different machine to the previous run. Paths are now relative to the package root rather than the machine root

### DIFF
--- a/__tests__/Loadable.test.js
+++ b/__tests__/Loadable.test.js
@@ -126,7 +126,12 @@ test("server side rendering flushing", async () => {
     return Loadable({
       loader: createLoader(400, null, new Error("test error")),
       LoadingComponent: MyLoadingComponent,
-      serverSideRequirePath: path.join(__dirname, "..", "__fixtures__", name)
+      serverSideRequirePath: path.join(
+        path.basename(__dirname),
+        "..",
+        "__fixtures__",
+        name
+      )
     });
   };
 
--- a/__tests__/__snapshots__/Loadable.test.js.snap
+++ b/__tests__/__snapshots__/Loadable.test.js.snap
@@ -3,35 +3,35 @@
 exports[`loading error 1`] = `
 <div>
   MyLoadingComponent 
-  {"isLoading":true,"pastDelay":false,"error":null}
+  {"prop":"baz","isLoading":true,"pastDelay":false,"error":null}
 </div>
 `;
 
 exports[`loading error 2`] = `
 <div>
   MyLoadingComponent 
-  {"isLoading":true,"pastDelay":true,"error":null}
+  {"prop":"baz","isLoading":true,"pastDelay":true,"error":null}
 </div>
 `;
 
 exports[`loading error 3`] = `
 <div>
   MyLoadingComponent 
-  {"isLoading":false,"pastDelay":false,"error":{}}
+  {"prop":"baz","isLoading":false,"pastDelay":false,"error":{}}
 </div>
 `;
 
 exports[`loading success 1`] = `
 <div>
   MyLoadingComponent 
-  {"isLoading":true,"pastDelay":false,"error":null}
+  {"prop":"foo","isLoading":true,"pastDelay":false,"error":null}
 </div>
 `;
 
 exports[`loading success 2`] = `
 <div>
   MyLoadingComponent 
-  {"isLoading":true,"pastDelay":true,"error":null}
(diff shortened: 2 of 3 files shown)


thejameskyle:
You can simply wrap `Loadable()` in your own higher-order component

alechill:
As explained I have tried wrapping an HOC over Loadable, but because LoadingComponent has to be passed by class reference rather than component instance there doesn't appear to be any chance to access the props of the Loading instance.

If you know of a way, then any pointers will be much appreciated

thejameskyle:
It's definitely possible, it's just composing functions. If you want to add something to make it more convenient I'm fine with that, but I disagree that this should be default behavior as depending on `props` like that creates implicit API contracts.

alechill:
Took a lot of wrangling but got a generic version working. Using a functional HOC caused re-render issues, but a more involved class based version that caches the wrapped components for subsequent updates seems to do the trick

```Typescript
import * as React from 'react'
import {CardContentLoading} from '../../CardContentLoading/CardContentLoading'
import {ICardContentExpanderProps} from '../../CardContentExpander/CardContentExpander'
import * as _ from './MapCardContent'

type LazyProps = _.IMapCardContentProps
type LazyComponent = typeof _.MapCardContent

import Loadable, {LoadableComponent, LoadingComponentProps, LoadedComponent} from 'react-loadable'

interface System {
  import<T> (module: string): Promise<T>
}
declare var System: System

export const lazy = <L, P extends L, T extends LoadedComponent<P>>
  (modulePath: string, loadingClass: React.ComponentClass<L> | React.SFC<L>)
    : React.ComponentClass<P> => {

  return class PurelyLazy extends React.PureComponent<P, void> {

    private Loading: React.StatelessComponent<LoadingComponentProps>
    private Lazy: LoadedComponent<P>

    constructor(props: P) {
      super(props)
      const UnderlyingLoading = loadingClass // react assumes lowercase is intrinsic component, must UC it locally
      this.Loading = (loadingProps: LoadingComponentProps) => <UnderlyingLoading {...props} {...loadingProps} />
      this.Lazy = Loadable<P, T>({
        loader: (): Promise<T> => System.import<T>(/* webpackChunkName: "lazy/[request]" */ `${modulePath}`),
        LoadingComponent: this.Loading
      })
    }

    render() {
      return <this.Lazy {...this.props} />
    }

  }

}

export default lazy<ICardContentExpanderProps, LazyProps, LazyComponent>('MapCardContent', CardContentLoading)
```

Ta for the lib BTW!

PR #43, opened 2017-05-29
Adds test case for desired behavior of babel plugin for user-space HOCs

lourd:
I discovered that when defining and using my own higher-order component in order to have a default `LoadingComponent`, the babel plugin does not work. It fails with a `TypeError`, `Cannot read property 'name' of undefined`.

This PR adds a failing test case demonstrating the behavior, and a snapshot of what I would expect the outcome to be. I understand the changes to the plugin necessary to produce that outcome would be pretty significant. If it's not possible, a statement should be added to the README about it not being a supported use case. You wouldn't think so right now, with the documentation of [how to not repeat](https://github.com/thejameskyle/react-loadable#how-do-i-avoid-repetition) coming right after a recommendation to use the Babel plugin.

### DIFF
--- a/__tests__/__snapshots__/babel-plugin.test.js.snap
+++ b/__tests__/__snapshots__/babel-plugin.test.js.snap
@@ -60,3 +60,21 @@ let C2 = Loadable({
   loader: () => import('./MyComponent')
 });"
 `;
+
+exports[`add-options-to-user-defined-HOC 1`] = `
+"import Loadable from 'react-loadable';
+
+function MyLoadable(opts) {
+  const LoadableComponent = Loadable({
+    LoadingComponent,
+    ...opts,
+  });
+  return props => <LoadableComponent {...props} />;
+}
+
+let C1 = MyLoadable({
+  loader: () => import('./MyComponent'),
+  serverSideRequirePath: _path.join(__dirname, './MyComponent'),
+  webpackRequireWeakId: () => require.resolveWeak('./MyComponent'),
+});"
+`;
--- a/__tests__/babel-plugin.test.js
+++ b/__tests__/babel-plugin.test.js
@@ -8,84 +8,78 @@ let trim = str => {
   return str.replace(/\r\n?/g, "\n").trim();
 };
 
-let fn = (code, opts = {}) => {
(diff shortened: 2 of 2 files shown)


thejameskyle:
The Babel plugin came later, there's no clear way on how to make it work with HOCs. My only idea is to do something like this:

```js
import Loadable, {loader} from 'react-loadable';
import LoadingComponent from './LoadingComponent';

function MyLoadable(opts) {
  return Loadable({
    LoadingComponent,
    ...opts,
  });
}

let MyLoadableComponent = MyLoadable({
  [loader]: () => import('./MyComponent'),
});
```

Where `loader` is a well-known symbol and the Babel plugin could work off of it.

lourd:
Not a bad idea. I could jibe with that API. If I write it, will you merge it?

faceyspacey:
@lourd perhaps with the Next.js style 2 argument API it might be easier: 

lourd:
Nice writeup! You've been crushing it 👊🏼

> Next.js style 2 argument API

I don't understand what you're referring to

faceyspacey:
Their dynamic function async takes an async component and options whereas this just takes options:

dynamic(asyncComponent, options).

So the first arg is always what was assigned to the loader key in React Loadable.

...on a side note, u have seen webpack's new magic comment feature right? That along with dynamic require paths allow U to define one hoc that dynamically chooses between multiple chunks/modules idiomatically and with one function. That will crush it for the HoC thing. There's actually a related issue I submitted to webpack or I would have wrote about it. resolveWeak doesn't yet fully work with dynamic module paths. Once we have his capability, the hoc stuff ur after will become even more important. In fact if ur like me, if ur code splitting, it almost always is done where u dynamically choose between a bunch of modules in a given category to get the most bang for your buck. Fortunately sokra marked it with high priority so I have a good feeling this will get done soon. They too have invested a lot in code splitting and likely get the importance of the path React Loadable pioneered.

Here's what needs to be fixed:
https://github.com/webpack/webpack/issues/4993

thejameskyle:
The [new Babel plugin](https://github.com/thejameskyle/react-loadable#babel-plugin-import-inspector) handles this.

PR #44, opened 2017-05-30
Parallel requests when importing

PerrineBoissieres:
This idea came when I ran into the necessity of loading localization files alongside chunks in a project.

You can use the new `parallels` option to send an array of promises to execute alongside the `import()`, and get the results in an `onLoaded` hook. The whole flow stays asynchronous.

```js
// example with fetch
const MyLoadable = Loadable({
  loader: () => import('MyComponent'),
  parallels: [
    () => fetch('path_to_resource').then(response => response.json()),
  ],
  onLoaded: ([module, data]) => {
    // store data, etc...
    // then we send the module so the flow can continue
    return Promise.resolve(module)
  },
})
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,11 @@ let LoadableMyComponent = Loadable({
   delay: 200,
   serverSideRequirePath: path.join(__dirname, './MyComponent'),
   webpackRequireWeakId: () => require.resolveWeak('./MyComponent'),
+  parallels: [fetch(), fetchNumber2()],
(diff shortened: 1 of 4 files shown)


thejameskyle:
I believe you can just do this:

```js
const MyLoadable = Loadable({
  loader: () => Promise.all([
    import('MyComponent'),
    fetch('path_to_resource').then(response => response.json()),
  ]),
  // ...
});
```

PR #45, opened 2017-06-02
Provide props to LoadingComponent

michalkvasnicak:
Provides `props` passed to wrapped component to `LoadingComponent` too.

Use case:

```js
const Loader = ({
  error,
  isLoading,
  onClose,
  pastDelay,
}: {
  error: ?Error,
  isLoading: boolean,
  pastDelay: boolean,
}) => {
  if (isLoading) {
    return pastDelay
      ? <Modal closable onClose={onClose}>
          <Loading size="large" />
        </Modal>
      : null;
  } else if (error) {
    return <Modal closable onClose={onClose}>Error</Modal>;
  }
  return null;
};

const AddressModal = Loadable({
  loader: () => import('../AddressModal'),
  LoadingComponent: Loader,
});

<AddressModal onClose={this.handleClose} />
```

### DIFF
--- a/__tests__/__snapshots__/Loadable.test.js.snap
+++ b/__tests__/__snapshots__/Loadable.test.js.snap
@@ -3,35 +3,35 @@
 exports[`loading error 1`] = `
 <div>
   MyLoadingComponent 
-  {"isLoading":true,"pastDelay":false,"error":null}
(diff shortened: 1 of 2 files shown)


thejameskyle:
See other PRs related to passing props through

PR #52, opened 2017-07-04
Add babel-plugin-transform-object-assign.

shamenchens:
It would be helpful for those runtime don't have `Object.assign` natively.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -4,5 +4,8 @@
       "loose": true
     }],
     "react"
+  ],
+  "plugins": [
+    "transform-object-assign"
   ]
 }
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
   },
   "devDependencies": {
     "babel-cli": "^6.24.1",
+    "babel-plugin-transform-object-assign": "^6.22.0",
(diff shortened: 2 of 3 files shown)


sibelius:
this polyfill should be in the client code

anilanar:
You can do that in your own project by transpiling `react-loadable/lib/index.js` using your own `.babelrc`. 

I would rather have ES2017 exported from `react-loadable`. Otherwise it's punishing (no matter how close this lib is) those who make effort to serve modern JS to modern browsers and legacy JS to legacy browsers.

👎

shamenchens:
Yes transpiling or use polyfill are options, but shouldn't `Object.assign` be listed on the dependencies if `react-loadable` didn't do that? IMO, as a lib, it could either add this object assign transform plugin or update the README.md that developers should take care transpiling themselves.

anilanar:
Adding to README is a better option IMO.

roderickhsiao:
Adding into README.md seems more reasonable as client might use different approach to polyfill `Object.assign`, which might causes duplication or unnecessary client side bundle increase.

PR #53, opened 2017-07-06
Add support for default exports

misantronic:
When working with typescript `import Loadable from 'react-loadable';` is not working.
`import * as Loadable from 'react-loadable';` on the other hand works, but doesn't match the default Loadable-typings.

So adding a default export to support both `import Loadable from 'react-loadable';` and typings would be neat.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -228,4 +228,6 @@ function LoadableMap(opts) {
 
 Loadable.Map = LoadableMap;
 
+Loadable.default = Loadable;
+
 module.exports = Loadable;


thejameskyle:
TypeScript is broken if they don't interop `module.exports` properly.

jimthedev:
@misantronic Any chance you forked the types or anything? I was able to resolve this just using require since I am using webpack but I was hoping to not need to do that. Ideally typescript would handle but of course at the moment it does not.

BeeeQueue:
@jamiebuilds Any chance to open/merge this? It's impossible to import this package if you have the tsconfig option `esModuleInterop` enabled.

PR #57, opened 2017-07-14
Fix undefined `error` being set

KenanY:
The `error` variable is undefined here, so this attempt to assign `err`
to it results in a `ReferenceError`. From the rest of the code, it looks
like this should have been `state.error`.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -86,7 +86,7 @@ function loadMap(obj) {
       });
     });
   } catch (err) {
-    error = err;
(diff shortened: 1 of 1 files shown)


loick:
Hi there, any ETA for this PR? :) 

Thanks!

tajo:
Thanks! Published as `v4.0.5`.

loick:
Thanks a lot!

PR #59, opened 2017-08-01
upgrade dependency with a fix of const usage

sejoker:
- [PR with a fix](https://github.com/thejameskyle/webpack-require-weak/pull/1)
- will resolve issue for browsers without support of ES6 (Safari 9, IE 11)


tajo:
Thanks! Released as `v4.0.4`.

PR #64, opened 2017-08-29
Always import when using webpack server-side

peterthompson:
When using webpack to transpile server code we use webpack weak Ids to resolve dynamic imports.

Server will cache these imports so the second time a module is required it will not be reported by the import inspector.

This change ensure that these imports are always reported when running server code transpiled by webpack.


thejameskyle:

PR #67, opened 2017-10-06
Introduce Loadable.preloadAll

thejameskyle:
I've been planning on doing this for a little while. It greatly simplifies server-side rendering.

Getting all of your components to render server-side from the first run all you need to do is wrap your startup code with `Loadable.preloadAll().then(() => ...);`

```js
Loadable.preloadAll().then(() => {
  app.listen(3000, () => {
    console.log('Running on http://localhost:3000/');
  });
});
```

Now, React Loadable doesn't care at all how you load your components, as long as you're returning promises.


theasia-Ashish:
Getting error `UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): ReferenceError: window is not defined` if i import css on component that is wrapped by `Loadable` in route. if i remove `Loadable` or `css import` or `Loadable.preloadAll().then(...` all working fine with server-side rendering but its displaying `Loading...`

PR #68, opened 2017-10-15
fix preloadReady

bertho-zero:


thejameskyle:
How is this fixing it?

bertho-zero:
Without this fix I have this error:
image

And with this fix I have all the files that load, not just those concerned.

thejameskyle:
Can you post exactly where in the react-loadable code that error is thrown? I can't tell from the stack trace because line numbers are based on your bundle.

bertho-zero:
I do not have the error above, but on the other hand without this fix I have the following error:

```
Warning: Text content did not match. Server: "ChatSend" Client: "Loading"
```

Otherwise the rest works perfectly well, the reason why all my routes loaded is that all my `Loadable` were in the same file...

thejameskyle:
> the reason why all my routes loaded is that all my Loadable were in the same file...

It shouldn't matter if they are all declared in the same file, it should only matter if they get rendered

bertho-zero:
Now I do not have the same client-side rendering as server side, I have the loading components at first render in the client without this fix.

thejameskyle:
I'm not sure why because this change does the exact opposite of what should be done. The updated test case is now incorrect behavior. Without an example of what you're doing I'm going to assume it is something wrong at your end and this just happens to fix it even if it introduces bad behavior.

PR #70, opened 2017-10-17
Delay 0 should show loading state in the current event loop

wenboyu2:
When delay is set to 0, loading state should show up immediately. Currently showing the loading state is delayed until the next event loop due to setTimeout.


thejameskyle:
Just FYI your git config is messed up so your GitHub profile isn't tied to this commit. I'll give you a chance to update it if you want.

wenboyu2:
@thejameskyle thanks! I've fixed the author info.

PR #71, opened 2017-10-17
Filter extraneous files

vlad-zhukov:


thejameskyle:
Not this module's job, handle it yourself.

vlad-zhukov:
Are you suggesting to do it during runtime?

thejameskyle:
Yes

PR #73, opened 2017-10-18
Extract js & css files for manifest

oviava:
Webpack will emit 4 files with the same 'rawname' if you have sourcemaps & also chunking css: [file].js [file].js.map [cssfile].css [cssfile].css.map.

This would allow us to exclude maps and also get the css files

### DIFF
--- a/README.md
+++ b/README.md
@@ -550,16 +550,24 @@ We can then render these bundles into `<script>` tags in our HTML.
 
 ```js
 let bundles = getBundles(stats, modules);
+/*
+  [{ id, name, js, css}]
+*/
 
 res.send(`
   <!doctype html>
   <html lang="en">
-    <head>...</head>
+    <head>
+    ...
+    ${bundles.map(bundle => {
+      return `<link href="/dist/${bundle.css}" rel="stylesheet"/>`
+    }).join('\n')}
+    </head>
     <body>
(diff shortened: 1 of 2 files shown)


thejameskyle:
This doesn't make much sense, a bundle is either css or js. Also you can just as easily split these apart in your server code:

```js
let bundles = getBundles(stats, modules);

let styles = bundles.filter(bundle => bundle.file.endsWith('.css'));
let scripts = bundles.filter(bundle => bundle.file.endsWith('.js'));

res.send(`
  <!doctype html>
  <html lang="en">
    <head>
      ...
      ${styles.map(style => {
        return `<link href="/dist/${style.file}" rel="stylesheet"/>`
      }).join('\n')}
    </head>
    <body>
      <div id="app">${html}</div>
      <script src="/dist/main.js"></script>
      ${scripts.map(script => {
        return `<script src="/dist/${script.file}"></script>`
      }).join('\n')}
    </body>
  </html>
`);
```

oviava:
Hey, apologies, perhaps should have offered some context using `extract-css-chunks-webpack-plugin` we can get separate css 'modules' together with the js chunk. They will be emitted under the same `module.rawRequest`

oviava:
After further reading and investigation - this is kinda redundant. Sorry for waisting your time :/

PR #74, opened 2017-10-19
Depend on prop-types the way Facebook recommends

lydell:
https://github.com/facebook/prop-types#how-to-depend-on-this-package

> For libraries, we *also* recommend leaving it in `dependencies`:
>
> ```js
>   "dependencies": {
>     "prop-types": "^15.5.7"
>   },
>   "peerDependencies": {
>     "react": "^15.5.0"
>   }
> ```
>
> **Note:** there are known issues in versions before 15.5.7 so we
> recommend using it as the minimal version.
>
> Make sure that the version range uses a caret (`^`) and thus is broad
> enough for npm to efficiently deduplicate packages.

---

Motivation:

- react-loadable currently has prop-types as a peer dependency.
- My project uses react-loadable but not prop-types.
- When installing dependencies, `yarn` correctly gives this warning:

```
warning "react-loadable@5.3.0" has incorrect peer dependency "prop-types@*".
```

Today I finally felt like doing something about that warning. There are two ways I can go:

1. Add prop-types to my project dependencies, and document that we only install it as a peer dependency for react-loadable.

2. This PR :)

(My project works even though I haven’t explicitly installed prop-types because other React libraries happen to depend on it ;) )

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,9 @@
     "start": "yarn build && webpack && babel-node example/server.js",
     "prepublish": "yarn build"
   },
+  "dependencies": {
+    "prop-types": "^15.5.7"
+  },
   "devDependencies": {
     "babel-cli": "^6.24.1",
     "babel-loader": "^7.1.2",
@@ -35,7 +38,6 @@
     "webpack": "^3.6.0"
   },
   "peerDependencies": {
-    "react": "*",
-    "prop-types": "*"
+    "react": "*"
   }
 }
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "prepublish": "yarn build"
   },
   "dependencies": {
-    "prop-types": "^15.5.7"
+    "prop-types": "^15.5.0"
   },
   "devDependencies": {
     "babel-cli": "^6.24.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -17,6 +17,9 @@
     "start": "yarn build && webpack && babel-node example/server.js",
     "prepublish": "yarn build"
   },
+  "dependencies": {
+    "prop-types": "^15.5.7"
```
thejameskyle: Can you open up the range more to include React 14, 15, and 16?

lydell: Sure! I just followed Facebook's instructions blindly ;) Which range would you like here? Or is it part of my job to find out?

thejameskyle: `"^0.14.0 || ^15.0.0 || ^16.0.0"`

lydell: As far as I can tell, these are the published versions:

```
❯ npm show prop-types versions
 
[ '0.1.0',
  '0.2.0',
  '15.5.0-alpha.0',
  '15.5.0',
  '15.5.1',
  '15.5.2',
  '15.5.3',
  '15.5.4',
  '15.5.5',
  '15.5.6',
  '15.5.7-alpha',
  '15.5.7-alpha.1',
  '15.5.7',
  '15.5.8',
  '15.5.9',
  '15.5.10',
  '15.6.0' ]
```

thejameskyle: Ah, figured prop-types was being kept in sync with react

PR #75, opened 2017-10-22
Refactor into bolt repo

jamiebuilds:
WIP

- Flow
- Bolt
- ES Modules


packages/react-loadable/index.js
```diff
@@ -0,0 +1,14 @@
+// @flow
```
Andarist: is this some leftover file? it seem that `packages/react-loadable/src/index.js` is the new entry point - and its exports are structured more nicely, reexporting things rather than assigning to the `Loadable`

also cant see `./lib/cjs` anywhere, nor any build script for it

thejameskyle: I'm not done

mvestergaard:
What is the status of this? Is there anything we can do to help?

This library is difficult to use in typescript due to how default exports are handled. And due to this currently not exposing an ES6 default export typescript gets confused. See https://github.com/DefinitelyTyped/DefinitelyTyped/issues/19337

nandosangenetto:
Sad, this branch is not working :(

PR #82, opened 2017-11-08
Add retry() function to the loading component when import failed

sanderdebruijne:
To improve the user experience, give the user a chance to retry loading the component when the import failed. The loading component receive a `retry` prop, which is a function to retry the operation.

```js
const Loading = props => {
  if (props.error) {
    return <div>Error! <button type="button" onClick={ props.retry }>Retry</button></div>;
  } else {
    return <div>Loading...</div>;
  }
};
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -215,7 +215,7 @@ will be `false`).
 ```js
 function Loading(props) {
   if (props.error) {
-    return <div>Error! <button type="button" onClick={ props.retry }>Retry</button></div>;
+    return <div>Error! <button onClick={ props.retry }>Retry</button></div>;
   } else {
     return <div>Loading...</div>;
   }
@@ -238,7 +238,7 @@ which will only be true once the component has taken longer to load than a set
 ```js
 function Loading(props) {
   if (props.error) {
-    return <div>Error! <button type="button" onClick={ props.retry }>Retry</button></div>;
+    return <div>Error! <button onClick={ props.retry }>Retry</button></div>;
   } else if (props.pastDelay) {
     return <div>Loading...</div>;
   } else {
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -210,13 +213,20 @@ function createLoadableComponent(loadFn, options) {
...
+
     render() {
       if (this.state.loading || this.state.error) {
         return React.createElement(opts.loading, {
           isLoading: this.state.loading,
           pastDelay: this.state.pastDelay,
           timedOut: this.state.timedOut,
-          error: this.state.error
+          error: this.state.error,
+          retry: this.retry.bind(this)
```
thejameskyle: Can you use a class property for `retry()` instead of using `.bind(this)`?

```js
retry = () => {
  // ...
};
```

README.md
```diff
@@ -831,10 +831,10 @@ This is the component you pass to [`opts.loading`](#optsloading).
 function LoadingComponent(props) {
   if (props.error) {
     // When the loader has errored
-    return <div>Error!</div>;
+    return <div>Error! <button type="button" onClick={ props.retry }>Retry</button></div>;
```
thejameskyle: You don't need `type="button"`

Flydiverny:
Think it would be nice to have this feature! Not sure how one would handle failing components currently. For example if an route which is react loadable fails to load it doesn't seem to retry even when navigating back and forth, as it uses the same loadable which is in the error state with no option to retry.

Flydiverny:
@thejameskyle Is there any discussion or thoughts on this somewhere? I saw you had some [tweet](https://twitter.com/thejameskyle/status/929944187914424320) in regards to naming of retry.

9teen90nine:
That would be really nice to have! Sometimes the chunk file loading gets interrupted and the corrupted file gets cached by browser, forcing user to clear the cache.

ifokeev:
UP Same issue here

delijah:
Can i help anything to get this merged into master?

Sigura:
@thejameskyle I will glad to help to add this feature to master too if that needed. That is really must have feature. How can I help to add it to npm asap?

sanderdebruijne:
Is it possible to get it merged or does it need some tweaks? If the internet connection of a user fails, an import fails, and it fails forever. If the user returns from within the same SPA-instance, it keeps failing, because the status won't be reset (it isn't even retried). Once failed, forever failed (per SPA-instance). In the current master branch, the only workaround is to completely refresh the page. I hope this pull request can be accepted to improve this user experience.

rubencodes:
Any update on this @jamiebuilds? This is a huge problem on mobile with unreliable connections right now.

PR #84, opened 2017-11-11
build manifest with webpack publicPath

bertho-zero:


src/webpack.js
```diff
@@ -16,7 +16,9 @@ function buildManifest(compiler, compilation) {
           manifest[module.rawRequest] = [];
         }
 
-        manifest[module.rawRequest].push({ id, name, file });
+        const path = compilation.outputOptions.publicPath || '';
+
+        manifest[module.rawRequest].push({ id, name, file: path + file });
```
thejameskyle: Hmm... I'm hesitant to make this change, I'd rather:

```js
const path = require('path');

// ...

let publicPath = path.join(compilation.outputOptions.publicPath || '', file);

manifest[module.rawRequest].push({ id, name, file, publicPath });
```

Then update the documentation accordingly.

bertho-zero: Having deleted my fork, I had to reopen a PR here: https://github.com/thejameskyle/react-loadable/pull/91

PR #86, opened 2017-11-15
Fixing Example TypeError

hutchgrant:
> TypeError: bundle.endsWith is not a function

should be: bundle.**file**.endsWith()


hutchgrant:
Yikes, outdated node, this runs fine without this fix.

PR #88, opened 2017-11-23
[WIP] pass params to loader (based on props)

debelbot:
Sometimes I've found I need to pass dynamic parameters to the loader, usually based on the props. This is useful because our component server can attach additional data to the component props.

If you agree with the general idea behind this PR, I'll be happy to polish it by writing tests and docs :)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -18,8 +18,8 @@ function isWebpackReady(getModuleIds) {
   });
 }
 
-function load(loader) {
-  let promise = loader();
+function load(loader, params) {
+  let promise = loader(params);
 
   let state = {
     loading: true,
@@ -104,13 +104,14 @@ function createLoadableComponent(loadFn, options) {
     render: render,
     webpack: null,
     modules: null,
+    propsToLoaderParams: props => props
   }, options);
 
   let res = null;
 
-  function init() {
+  function init(params) {
     if (!res) {
-      res = loadFn(opts.loader);
+      res = loadFn(opts.loader, params);
     }
     return res.promise;
   }
@@ -128,7 +129,7 @@ function createLoadableComponent(loadFn, options) {
   return class LoadableComponent extends React.Component {
     constructor(props) {
       super(props);
-      init();
+      init(opts.propsToLoaderParams(props));
 
       this.state = {
         error: res.error,
@@ -145,8 +146,8 @@ function createLoadableComponent(loadFn, options) {
       }),
     };
 
-    static preload() {
-      return init();
+    static preload(params) {
+      return init(params);
     }
 
     componentWillMount() {


diruuu:
I need this.

thejameskyle:
I'm not a fan of this. Your loader should be as static as possible or it's impossible to do things like server-side rendering reliably. I know that it is very limiting, but it's on purpose.

dyst5422:
How then should we do something dynamic following the loading of a bundle?

ie replace redux reducer with new reducer from bundle prior to calling the render

thejameskyle:
I don't know what that means but you can probably do it by overriding `opts.render`

```js
  loader: () => import('./Foo'),
  loading: Loading,
  render(loaded, props) {
    return <loaded.default {...props}>;
  } 
});
```

thejameskyle:
Why not just:

```js
const FancyHello = Loadable({ loader: () => import('./FancyHello'), ... });
const RegularHello = Loadable({ loader: () => import('./RegularHello'), ... });

const Hello = props =>
  props.name === 'Bob'
    ? <FancyHello {...props}/>
    : <RegularHello {...props}/>;
```

Doing a bunch of dynamic stuff within `import(...)` isn't even really supported by bundlers.

thejameskyle:
I'm not interested in supporting that use case in this way

dyst5422:
The case I am thinking about is like so

```ts
const Hello = Loadable({
  loader: (props) => import('./someModule').then(bundle => {
    props.reduxStore.replaceReducers(bundle.reducer);
    return bundle.component;
  },
  ...
});
```

It occurs to me now though that this sort of logic should probably live in `componentWillMount()`

thejameskyle:
```js
const Hello = Loadable({
  loader: () => import('./someModule'),
  render(loaded, props) {
    props.reduxStore.replaceReducers(loaded.reducer);
    return loaded.component;
  }
});
```

dyst5422:
That doesn't work because switching the reducers triggers a render...within a render. 

Definitely feel that `componentWillMount()` is the place for that kind of thing though.

```ts
const Hello = Loadable({
  loader: () => import('./someModule'),
  render(loaded, props) {
    props.reduxStore.replaceReducers(loaded.reducer);
    return <loaded.component reducer={loaded.reducer} {...props}/>;
  }
});

/*******************************/

class MyComponent extends React.Component {
  componentWillMount(props) {
    props.reduxStore.replaceReducers(props.reducer);
  }
...
}
```

maubrsp:
@debelbot 

I think something like this may help

const Hello = props => {
 let Loader =Loadable({
  loader: () => import(props.name === 'Bob' ? './FancyHello' : './RegularHello'),
  ...
});
return (<Loader />)
}


 <Hello props={ something } />

dbbk:
This would be useful for us! I just want to be able to retrieve routing params in the loader. Here's our code:

```js
loader: props => import('../components/AsyncComponent').then(module => {
  const ModuleComponent = module.default;

  if ('getInitialProps' in ModuleComponent) {
    return ModuleComponent.getInitialProps(props).then(resolvedProps => {
      return <ModuleComponent {...resolvedProps} />;
    });
  }

  return <ModuleComponent />;
}),
```

If there's a better way of accomplishing this I'd be grateful for suggestions.

thejameskyle:
```js
Loadable({
  loader: () => import('../components/AsyncComponent'),
  render(module, props) {
    const ModuleComponent = module.default;

    if ('getInitialProps' in ModuleComponent) {
      return ModuleComponent.getInitialProps(props).then(resolvedProps => {
        return <ModuleComponent {...resolvedProps} />;
      });
    }

    return <ModuleComponent />;
  }
});
```

dbbk:
@thejameskyle I don't think that'll work as `Loadable.render` won't allow you to return a promise?

dekryptic:
@jamiebuilds Are you returning a `Promise` in render inside the if-statement? Is that permitted?

jaybe78:
@dekryptic @dbbk @debelbot @maubrsp @dyst5422 @diruuu 
Hi Guys, I guess you've probably forked the project by now, and modify the repo according to your needs...
In al cases, may be we could think of a "good" solution to provided that functionality without hindering SSR.
Apparently the project has been abandoned, so I've forked the project and merged most important PRs there (migration to babel 7, webpack 4,...)
Check this out 
https://github.com/jaybe78/loadable-example/
https://www.npmjs.com/package/jaybe-react-loadable

jaybe78:
@debel Yep right! does not support SSR yet though.

PR #90, opened 2017-11-25
Fix bundle filters to match documentation

chrisself:
Currently, the server side rendering example results in the following error:
```
TypeError: bundle.endsWith is not a function
```
This pull request updates the bundle filtering to match the provided documentation.


thejameskyle:
Updated in other PR

PR #91, opened 2017-11-27
build manifest with webpack publicPath

bertho-zero:


bertho-zero:
@thejameskyle it's related to https://github.com/thejameskyle/react-loadable/pull/84

thejameskyle:
Why was `yarn.lock` updated?

bertho-zero:
This is because I made a recent installation (`yarn`). If you prefer I can cancel?

thejameskyle:
Yarn should not have rewritten the yarn.lock file so dramatically unless you deleted it or tried to change dependencies. Can you remove it from your commit?

bertho-zero:
I removed it form the commit.

PR #93, opened 2017-11-30
publicPath could be an url

fdubost:
Use of `url.resolve` instead of `path.join` because `publicPath` could be an url with host domain (particularly in dev mode).


fdubost:
Could you please publish a new tag on npm ? 😉

PR #94, opened 2017-12-09
compatibility with ConcatenatedModule

salzhrani:
when using `webpack.optimize.ModuleConcatenationPlugin` some modules are instances of `ConcatenatedModule` which have no `rawRequest` property, which breaks the manifest.
I tried this change with the example in the repo and it worked fine.


InsaneViku:
I think this would be great to get in, especially to take advantage of this webpack optimization alongside `react-loadable` (which is addressing all my code-splitting needs btw). Any plans to merge this in anytime soon, @thejameskyle?

PR #95, opened 2018-01-02
add install command to readme

pravdomil:
...


pravdomil:
npm is not allowed right?

PR #99, opened 2018-01-19
Allow to display loading and loaded Component at the same time

danielberndt:
Think of a scenario where you're showing a loading spinner that you want to fade out once the content is loaded.
This would require the loading and the loaded component to be shown simultaneously. Something that the current `render` method of the `LoadableComponent` does not support.

This is why I'm suggesting to add a `custom` option to `Loadable` that gives the user full control over what is rendered. This would allow code like this:

```jsx
Loadable({
  loader: () => import("./MyComponent"),
  custom: ({isLoading, loaded, props}) => (
    <Motion style={{opacity: spring(isLoading ? 1 : 0)}}>
      {({opacity}) => (
        <div>
          {opacity !== 0 && <Spinner opacity={opacity}/>}
          {loaded && React.createElement(loaded, {opacity: 1 - opacity, ...props)}
        </div>
      )}
    </Motion>
  )
})
```

Please let me know how you feel about this change and I'm happy to add changes to the docs as well :)

### DIFF
--- a/__tests__/__snapshots__/test.js.snap
+++ b/__tests__/__snapshots__/test.js.snap
@@ -1,5 +1,25 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`custom render 1`] = `
+<div>
+  <div>
+    loading...
+  </div>
+</div>
+`;
+
+exports[`custom render 2`] = `
+<div>
+  <div>
+    done loading, we could fade this out now
+  </div>
+  <div>
+    MyComponent 
+    {"prop":"foo"}
+  </div>
+</div>
+`;
+
 exports[`delay and timeout 1`] = `
 <div>
   MyLoadingComponent 
--- a/__tests__/test.js
+++ b/__tests__/test.js
@@ -151,6 +151,24 @@ test('render', async () => {
   expect(component.toJSON()).toMatchSnapshot(); // success
 });
 
+test('custom render', async () => {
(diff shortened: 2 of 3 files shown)


thejameskyle:
I'm not going to add another method, I think you can get this working with `opts.render` if you just add the state as a third argument

danielberndt:
Thanks for the pointer @thejameskyle!

I just got it working with the existing API.
It's a bit weird - but not that much. And I think it's better than bloating the api.

Here's the code (and the corresponding [codesandbox](https://codesandbox.io/s/xln1ylovp4)):

```jsx
class Loading extends React.Component {
  render() {
    const { children } = this.props;
    return (
      <Motion style={{ isReady: spring(children ? 1 : 0) }}>
        {({ isReady }) => (
          <div>
            <div style={{opacity: 1 - isReady}}>Loading</div>
            {children && <div style={{opacity: isReady}}>{children}</div>}
          </div>
        )}
      </Motion>
    );
  }
}

  loader: () => import("./Comp"),
  loading: Loading,
  render(loaded, props) {
    return <Loading>{React.createElement(loaded.default, props)}</Loading>;
  }
});

class App extends React.Component {
  render() {
    return <LoadableComponent />;
  }
}
```

Well, I guess I got what I wanted :)
Do you think it's worth pointing out this approach in the docs? I feel that being able to animate this kind of thing could benefit some people.

PR #105, opened 2018-02-14
more global pre-loaded components resolver

berkaey:
Hi there,  when you try to use SSR and code-splitting without webpack (like in meteor apps) react-loadable only checks for webpack. 

SSR works but we've got Hydration Warnings because some modules were loaded already.

I've added moduleId based checking and all HydrationWarnings disappeared, now on your first render you can preload modules from backend again to sync with client (or on some other platform you may bundle loaded components to client also).

meteor example

/imports/startup/server.js
```
onPageLoad(async sink => {

  let modules = [];

  const { locale } = sink.request.cookies;

  const App = props => (
    <Loadable.Capture
      report={mod => {
        modules.push(mod);
      }}
    >
      <Routes {...props} locale={locale} />
    </Loadable.Capture>
  );

  await Loadable.preloadAll();

  const html = renderToString(<App location={sink.request.url} />)
  sink.renderIntoElementById("app", html);

  sink.appendToBody(`
    <script>
      window.__PRELOADED_COMPONENTS__ = ${JSON.stringify(modules)}
    </script>
  `);
});
```

/imports/startup/client.js
```
onPageLoad(async () => {
  await Loadable.preloadReady().then(() => {
    delete window.__PRELOADED_COMPONENTS__;
    ReactDOM.hydrate(<App />, document.getElementById("app"));
  });
});
```


jamiebuilds:
Please don't change all of the file permissions and run through your own code formatter. It makes it hard to know whats going on

PR #107, opened 2018-02-20
Improve rendering

adam-26:
@jamiebuilds, thanks for the great work.

I have a use-case not currently supported by `react-loadable`; to block rendering of deeply nested routes that use dynamically loaded components until all those components are loaded.

This PR enables `react-loadable` to be used for the above use-case. I've made a few small modifications to improve the ability to fully control the render lifecycle of dynamically loaded components. 

1. For custom render methods without a `loading` prop, pass `state` instead of `loaded` (Also fixes #99)
2. Allow developers to manually preload modules on the client
3. `render` also accepts react elements

Note: with these changes the `loading` prop becomes _optional_ as it is now possible to render custom loading content using the `render()` method.

I've commented each of the commits in the pull request, but I'll repeat those here for a quick summary.

#### Commit 1
When a `loading` prop is defined on a `Loadable`, the behavior remains backwards compatible. When **no** `loading` prop is defined, the `render(state, props)` method receives a loadable `state` argument instead of `loaded` argument.
~~Instead of only passing `loaded` to custom `render()`, pass _state_ to give full control when rendering the dynamically loaded component(s).~~ This allows easily implementing behaviors such as transitions, and blocking of nested route rendering until all modules are loaded.

This is ~~_almost_~~ entirely backwards compatible. ~~The **only** change required by users of previous versions is to use `object destructuring` syntax for the “loaded” argument. This can be done with a simple find & replace.~~

~~Example: `render(loaded, props)` must change to `render({ loaded }, props)`~~

#### Commit 2
Allows developers to manually preload webpack modules.

**use-case**: _preloading all modules required by a deeply nested route_
When dynamically loading a deeply nested route (many levels deep), you can provide a better user experience when all required modules are loaded before rendering the route for the user.
 * This prevents waterfall rendering of nested components (where a user can see parent components render before child components).
 * Allows developers to easily handle errors and display an error page if the route fails to load, rather than having to display error messages for each component/module that failed to load

Example:
```js

const LoadableContent = Loadable({
    loader: () => import('./PreLoadedContent')
});

// Use loadable to manually preload modules using 'loaders'
Loadable.preload([LoadableContent.getLoader()]).then(() => {
  this.setState({ isLoaded: true });
});

```
See the updated example in the PR for a better example.

#### Commit 4
The `render` property can also  accept a react element for rendering the loadable.
```js
function CodeSplitRenderer(props) {
  const { codeSplit, ...rest } = props;
  return <codeSplit.loaded {...rest} />;
}

const myLoadable = Loadable({
  render: <CodeSplitRenderer myProp="someValue" />
});
```

Let me know what you think, ~~I know the breaking change isn't ideal but its small and easy to fix for developers~~. I think these changes allow `react-loadable` to be used for a very common use-case for which there are many less-elegant custom solutions blogged about.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1222,6 +1222,12 @@ export default {
 This will create a file (`opts.filename`) which you can import to map modules
 to bundles.
 
+### `opts.filename`
+Required, the destination file for writing react-loadable module data
+
+### `opts.ignoreChunkNames`
+Optional, an array of webpack chunk names to exclude from the module data
+
 [Read more about mapping modules to bundles](#mapping-loaded-modules-to-bundles).
 
 ### `getBundles`
--- a/README.md
+++ b/README.md
@@ -305,6 +305,8 @@ Loadable({
 
 ### Customizing rendering
 
+`render(state, props)` can access `state` and `props` to determine the current state of the loader.
+
 By default `Loadable` will render the `default` export of the returned module.
 If you want to customize this behavior you can use the
 [`render` option](#optsrender).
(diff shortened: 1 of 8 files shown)


adam-26:
@jamiebuilds, just to let you know I've modified the PR so its fully backwards compatible.

PR #108, opened 2018-02-21
Inject props to `loading` component

ehellman:
To be able to affect the loading indication with more context.

For us, this would help because we already have a loading indicator for our XHR request, and this loading indicator is wrapping the feature that is being code split. So without this, there would be two loading indicators visible/used.

With this change, we can inject our own loading state into Loadable when needed.

// @ehellman & @danielwerthen


jamiebuilds:
Please look at other pull requests before opening new ones, this has been rejected before

PR #109, opened 2018-02-23
Update version to 5.3.2.

PaulJuliusMartinez:
There have been quite a few changes since the last bump in version, most notably https://github.com/jamiebuilds/react-loadable/commit/60495e387c11e97a11a3087402afa589d822c0fc which exposes the `publicPath` from the Webpack config.

I just want an updated version so I can use that feature, but I couldn't file an Issue.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-loadable",
-  "version": "5.3.1",
+  "version": "5.3.2",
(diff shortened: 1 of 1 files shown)


lucianohgo:
@jamiebuilds Could you provide an explanation why this can't be bumped?

lucasterra:
@jamiebuilds We need a release, please. :)

PR #111, opened 2018-03-07
Adds `webpackReport` prop/function

leebenson:
This PR adds a `webpackReport` function alongside the existing `report` function to `<Loadable.Capture>`, to return module IDs via calls to `require.resolveWeak()`

Unlike module names, IDs act as a unique identifier for chunks (note: I've tested this for Webpack 4; I'm not sure if the Webpack 3 stats API/JSON dump is different... _possibly_ due to the `SplitChunksPlugin` vs. `CommonChunks`, but I'm not sure.)

Assuming access to both client and server webpack stats in SSR middleware, this makes it trivial to cross-reference chunk identifiers and avoid edge cases where relative module names are identical, yet reference different chunks. An example might be having, say, a "./content" import which is different for every parent component; `report` can't distinguish between, but the above approach can.

Given SSR module capturing that looks something like this:

```jsx
const components = (
  <Loadable.Capture
    report={(moduleName: string) => modules.push(moduleName)}
    webpackReport={(moduleName: string) => webpackModules.push(moduleName)}>
    <StaticRouter location={ctx.request.url} context={routeContext}>
      <Root />
    </StaticRouter>
  </Loadable.Capture>
);
```

... the relevant client-side chunks can be located like this:

```ts
const chunks = [].concat(
  ...webpackModules
    .map(moduleId => (
      [].concat(
        ...clientStats.modules
        .find((m: any) => m.identifier === serverStats.modules[moduleId].identifier).chunks
        .filter((m: any) => m /* filter for truthy values */)
        .map((c: any) => clientStats.chunks[c].files)
      ) 
    ))
);

const js = chunks.filter((c: string) => c.endsWith(".js"));

console.log("js -> ", js); // <--- DUMP OUT THE .JS SCRIPTS
```

... resulting in something like:

> js ->  [ 'assets/js/1.eb4231d455a06b1e155c.js', 'assets/js/4.a60aba1bbe9940e49e91.js' ]

Which can then be included in the HTML render.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,8322 @@
+{
+  "name": "react-loadable",
+  "version": "5.3.1",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "abab": {
+      "version": "1.0.4",
+      "resolved": "https://registry.npmjs.org/abab/-/abab-1.0.4.tgz",
+      "integrity": "sha1-X6rZwsB/YN12dw9xzwJbYqY8/U4=",
+      "dev": true
+    },
+    "accepts": {
+      "version": "1.3.5",
+      "resolved": "https://registry.npmjs.org/accepts/-/accepts-1.3.5.tgz",
+      "integrity": "sha1-63d99gEXI6OxTopywIBcjoZ0a9I=",
+      "dev": true,
+      "requires": {
+        "mime-types": "2.1.18",
+        "negotiator": "0.6.1"
+      }
+    },
+    "acorn": {
+      "version": "4.0.13",
+      "resolved": "https://registry.npmjs.org/acorn/-/acorn-4.0.13.tgz",
+      "integrity": "sha1-EFSVrlNh1pe9GVyCUZLhrX8lN4c=",
+      "dev": true
+    },
+    "acorn-dynamic-import": {
+      "version": "2.0.2",
+      "resolved": "https://registry.npmjs.org/acorn-dynamic-import/-/acorn-dynamic-import-2.0.2.tgz",
+      "integrity": "sha1-x1K9IQvvZ5UBtsbLf8hPj0cVjMQ=",
+      "dev": true,
+      "requires": {
+        "acorn": "4.0.13"
+      }
+    },
+    "acorn-globals": {
+      "version": "3.1.0",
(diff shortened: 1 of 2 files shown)


jamiebuilds:
Could this be written as an additional argument to `report(moduleName, webpackId)`?

Also please remove the `package-lock.json` that was added in this PR

qq281113270:
> This PR adds a `webpackReport` function alongside the existing `report` function to `<Loadable.Capture>`, to return module IDs via calls to `require.resolveWeak()`
> 
> Unlike module names, IDs act as a unique identifier for chunks (note: I've tested this for Webpack 4; I'm not sure if the Webpack 3 stats API/JSON dump is different... _possibly_ due to the `SplitChunksPlugin` vs. `CommonChunks`, but I'm not sure.)
> 
> Assuming access to both client and server webpack stats in SSR middleware, this makes it trivial to cross-reference chunk identifiers and avoid edge cases where relative module names are identical, yet reference different chunks. An example might be having, say, a "./content" import which is different for every parent component; `report` can't distinguish between, but the above approach can.
> 
> Given SSR module capturing that looks something like this:
> 
> ```js
> const components = (
>   <Loadable.Capture
>     report={(moduleName: string) => modules.push(moduleName)}
>     webpackReport={(moduleName: string) => webpackModules.push(moduleName)}>
>     <StaticRouter location={ctx.request.url} context={routeContext}>
>       <Root />
>     </StaticRouter>
>   </Loadable.Capture>
> );
> ```
> 
> ... the relevant client-side chunks can be located like this:
> 
> ```ts
> const chunks = [].concat(
>   ...webpackModules
>     .map(moduleId => (
>       [].concat(
>         ...clientStats.modules
>         .find((m: any) => m.identifier === serverStats.modules[moduleId].identifier).chunks
>         .filter((m: any) => m /* filter for truthy values */)
>         .map((c: any) => clientStats.chunks[c].files)
>       ) 
>     ))
> );
> 
> const js = chunks.filter((c: string) => c.endsWith(".js"));
> 
> console.log("js -> ", js); // <--- DUMP OUT THE .JS SCRIPTS
> ```
> 
> ... resulting in something like:
> 
> > js ->  [ 'assets/js/1.eb4231d455a06b1e155c.js', 'assets/js/4.a60aba1bbe9940e49e91.js' ]
> 
> Which can then be included in the HTML render.

PR #112, opened 2018-03-08
Allow default options & loader fn to be passed straight to Loadable

sdgluck:
It gets a bit verbose if you have lots of `Loadable`s that only have `options{loader,loading}` where each share the same `loading` component.

This PR would allow the following...

```
// BEFORE
const Start = Loadable({
  loading: Loading,
  loader: () => import('../Home')
})

const User = Loadable({
  loading: Loading,
  loader: () => import('../User')
})

const Feed = Loadable({
  loading: Loading,
  loader: () => import('../Feed')
})

```

```
// AFTER
Loadable.setDefaultOptions({
  loading: Loading
})

const Start = Loadable(() => import('../Home'))
const User = Loadable(() => import('../User'))
const Feed = Loadable(() => import('../Feed'))
```

Just a suggestion in the form of a PR, as issues aren't open.

The README would need updating, which I will do if this functionality is desirable.

Thanks for a great lib!

### DIFF
--- a/__tests__/__snapshots__/test.js.snap
+++ b/__tests__/__snapshots__/test.js.snap
@@ -97,28 +97,56 @@ exports[`loading error 3`] = `
 </div>
 `;
 
-exports[`loading success 1`] = `
+exports[`loading success default options.loading, fn arg 1`] = `
 <div>
   MyLoadingComponent 
   {"isLoading":true,"pastDelay":false,"timedOut":false,"error":null}
 </div>
 `;
 
-exports[`loading success 2`] = `
+exports[`loading success default options.loading, fn arg 2`] = `
 <div>
   MyLoadingComponent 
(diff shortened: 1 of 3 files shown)


madiodio:
Isn't it what is specified in this [FAQ section](https://github.com/jamiebuilds/react-loadable#how-do-i-avoid-repetition)? Or I assume the following could be added to that section...
````js
import Loading from './my-loading-component';

const MyLoadable = loader => {
  try {
    return Loadable({
      loader,
      loading: Loading,
      delay: 300
    });
  } catch (error) {
    throw new Error(error);
  }
};

const Start = MyLoadable(() => import('../Home'));
const User = MyLoadable(() => import('../User'));
const Feed = MyLoadable(() => import('../Feed'));
````

sdgluck:
That does the trick. 😃 Thanks!

PR #113, opened 2018-03-28
pass in given props to loading component

ignatiusreza:
First of all, thanks for this package!

Now about the PR.. we want to be able to have a loading component that doesn't look like one.. for example, when loading:

- autocomplete or datepicker field can be displayed as a normal textfield 
- custom select box can be displayed as either a normal select or a button (depending on your design)
- element which display modal, tooltip or other type of popper on click/hover can be rendered with no interaction
- lower quality image while the original image is loading

to achieve this, we need to have access to the original given props when rendering the loading component, so that we can apply the `className` or read any `value` as required..

### DIFF
--- a/__tests__/__snapshots__/test.js.snap
+++ b/__tests__/__snapshots__/test.js.snap
@@ -3,21 +3,21 @@
 exports[`delay and timeout 1`] = `
 <div>
   MyLoadingComponent 
-  {"isLoading":true,"pastDelay":false,"timedOut":false,"error":null}
+  {"isLoading":true,"pastDelay":false,"timedOut":false,"error":null,"prop":"foo"}
 </div>
 `;
 
 exports[`delay and timeout 2`] = `
 <div>
   MyLoadingComponent 
-  {"isLoading":true,"pastDelay":true,"timedOut":false,"error":null}
+  {"isLoading":true,"pastDelay":true,"timedOut":false,"error":null,"prop":"foo"}
 </div>
 `;
 
 exports[`delay and timeout 3`] = `
 <div>
   MyLoadingComponent 
-  {"isLoading":true,"pastDelay":true,"timedOut":true,"error":null}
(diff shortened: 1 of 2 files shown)


jamiebuilds:
Please read all the other PRs that have been opened about this

ignatiusreza:
Hi Jamie,

Sorry, didn't noticed that there was other similar PRs opened.. now that I've read them, I assumed that you were thinking about the case in #88?

if yes, then the case that i'm trying to solve here is fundamentally different.. as mentioned in this PR description, we're not trying to do anything dynamic or complicated.. we just want to pass in a couple of standard props like `id`, `className`, 'value`, and `children` to the loading component..

this is necessary to allow for seamless transition between loading and loaded, e.g. an autocomplete field will by styled just like a field when loading and loaded.. this is important in term of giving user the best experience in using the site..

with that in mind, would you be open to passing down only a select view of whitelisted props? or do you have other suggestion on how to achieve the desired effect?

CurtisHumphrey:
I would suggest passing the original props on to the loader via something like props.original_props so as to prevent issues where a prop like error is pass both from the loader and from the original component

jamiebuilds:
> we just want to pass in a couple of standard props like id, className, value, and children to the loading component..

add them to the loading component directly

ignatiusreza:
thanks for the suggestion! but, that doesn't sounds like a sustainable solution.. for `className` maybe, but for `id`, `value` and `children`, it would mean that we would need to write a different loadable wrapper each time we want to use the same component..

for example, let's say that we have a component for handling phone input.. and within a single form, we ask guest to type in their home number and office number.. for the sake of the argument, let's say that value is stored in redux store and we're using react-redux-form to read from the store..

now, if we need to add `value` to each loading component for phone input individually, we would need to create 2 loadable wrapper, with each reading from different part of the redux store..

PR #115, opened 2018-04-05
Update README.md

budde377:
With WebPack 4 I didn't need to extract the manifest. It worked out of the box.


7rulnik:
@budde377 @tajo yeah, it works. But chunks will start loading after the application starts up.
So it's still make sense.

PR #116, opened 2018-04-09
add preFetch api

ColorfulCow:
A usage scenario: inside of component, i provide a static function named preFetch; in react-router's onEnter hooks, i can directly call static preFetch function to load asyn data before component render


jamiebuilds:
no thanks

dekryptic:
Where do you recommend someone use the current version of react-loadable to fetch data server-side? `Loadable.Map`?

PR #117, opened 2018-04-16
Add a preloadables preloader, which skips webpack checks for meteor, …

CaptainN:
I wanted to see if we could use this to do SSR with Meteor, but it turns out that SSR with React Loadable requires a webpack plugin. Meteor uses its own build system, so we can't use the webpack plugin. Meteor's dynamic import system is completely different from webpack anyway, in that it bundles are assembled at runtime based on individual requests for packages and files.

So I did a couple of things to make this work in Meteor - some of them are probably fine in general, but one change to the babel plugin may need to be better thought through.

What I did is to expose a new method on Loadable - `preloadablesReady`. I tried to keep everything in the spirit of the current API. This method works in practice a lot like `preloadReady` except it takes a list of modules to preload. This list is generated server side by the added `Capture.reportResolved` which will run the `webpack` method to get a translated rooted path (I didn't have any luck loading relative imports). The `preloadablesReady` method also skips the `isWebpackReady` method, since we are using Meteor and that check always fail.

Setting this up is pretty easy:

Server:
```JavaScript
import React from 'react'
import { StaticRouter } from 'react-router'
import { renderToString } from 'react-dom/server'
import { onPageLoad } from 'meteor/server-render'
import Loadable from 'react-loadable'
import App from '/imports/App'

Loadable.preloadAll().then(() => onPageLoad(sink => {
  const context = {}
  const modules = []
  const modulesResolved = []
  sink.renderIntoElementById('root', renderToString(
    <Loadable.Capture report={(moduleName) => { modules.push(moduleName) }}
      reportResolved={(resolvedModuleName) => { modulesResolved.push(resolvedModuleName) }}>
      <StaticRouter location={sink.request.url} context={context}>
        <App />
      </StaticRouter>
    </Loadable.Capture>
  ))
  sink.appendToBody(`<script>
  var preloadables = ${JSON.stringify(modulesResolved)}
</script>`)
```

Client:
```JavaScript
import { onPageLoad } from 'meteor/server-render'
import React from 'react'
import Loadable from 'react-loadable'
import { hydrate } from 'react-dom'
import { BrowserRouter } from 'react-router-dom'

onPageLoad(async sink => {
  let App = (await import('/imports/App')).default
  if (window.preloadables) {
    await Loadable.preloadablesReady(preloadables)
  }
  hydrate(
    <BrowserRouter><App /></BrowserRouter>,
    document.getElementById('root')
  )
})
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 .idea
 node_modules
 *.log
-lib
 coverage
 example/dist
--- a/lib/babel.js
+++ b/lib/babel.js
@@ -1,0 +1,73 @@
+'use strict';
+
+exports.__esModule = true;
+
+exports.default = function (_ref) {
+  var t = _ref.types,
+      template = _ref.template;
+
+  return {
+    visitor: {
+      ImportDeclaration: function ImportDeclaration(path) {
+        var source = path.node.source.value;
+        if (source !== 'react-loadable') return;
+
+        var defaultSpecifier = path.get('specifiers').find(function (specifier) {
+          return specifier.isImportDefaultSpecifier();
+        });
+
+        if (!defaultSpecifier) return;
+
+        var bindingName = defaultSpecifier.node.local.name;
+        var binding = path.scope.getBinding(bindingName);
+
+        binding.referencePaths.forEach(function (refPath) {
+          var callExpression = refPath.parentPath;
+
+          if (callExpression.isMemberExpression() && callExpression.node.computed === false && callExpression.get('property').isIdentifier({ name: 'Map' })) {
+            callExpression = callExpression.parentPath;
+          }
+
+          if (!callExpression.isCallExpression()) return;
+
+          var args = callExpression.get('arguments');
+          if (args.length !== 1) throw callExpression.error;
+
+          var options = args[0];
+          if (!options.isObjectExpression()) return;
+
+          var properties = options.get('properties');
+          var propertiesMap = {};
+
+          properties.forEach(function (property) {
+            var key = property.get('key');
+            propertiesMap[key.node.name] = property;
+          });
+
+          if (propertiesMap.webpack) {
+            return;
+          }
+
+          var loaderMethod = propertiesMap.loader.get('value');
+          var dynamicImports = [];
+
+          loaderMethod.traverse({
+            Import: function Import(path) {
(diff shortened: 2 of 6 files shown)


CaptainN:
RL's preferred way to go seems to be to hide some of the complexity by using hidden state. This is cool, if a bit hard to extend from the outside (I tried everything to get this to load without having to modify RL). There may then be a better way to capture the resolved moduleNames, and send them into the `preloadablesReady` method (or a least a way that's more consistent with the way it works with WebPack, where that detail is hidden). I also just realized I haven't tested the way I'm pre-resolving the modules with npm modules - I don't know if it'll work there.

There are some other changes that might make sense - like making `reportResolved` more of a flag, which effects the output of the normal (required) report method on `Capture`.

Anyway, I didn't want to spend too much more time on this until getting some feedback. Cheers!

ardatan:
You can use `webpack` with Meteor
https://github.com/ardatan/meteor-webpack

CaptainN:
That's true, but it's not the meteor way - and I don't think it would solve the problem in this case anyway. Meteor delivers missing modules at runtime based on a requests from the client (which only requests missing modules). In Webpack modules are encapsulated into bundles along code split boundaries. There's a bunch of stuff in the webpack plugin to map modules to bundles, and we don't need any of that in Meteor.

There's also a compatibility change I made in this PR (`require.resolveWeak` to `require.resolve`) - Meteor doesn't support `resolveWeak`. I have no idea what the difference is, since a few google searches hadn't turned up much in the way of documentation for resovleWeak.

CaptainN:
I ended up forking this to create a Meteor centric version. In the process I also ported the component to use hooks, to fix up some warnings about deprecated use of `componentWillMount`. Since it's now completely different, I'll go ahead and close this PR. Thanks for providing a great base for me to build upon.

CaptainN:
BTW, you can find the fork I made here: https://github.com/CaptainN/npdev-react-loadable

It contains a bunch of changes to make it compatible with Meteor, but also includes a port from the old class based component with `componentWillMount` (which kicks up a warning as of React 16.9) to a [hooks based implementation](https://github.com/CaptainN/npdev-react-loadable/blob/6b2235b1aad266eacd61191bb823b68d39e2ee5f/react-loadable-client.js#L64) that I thought you may find useful.

PR #118, opened 2018-04-17
Added props to loading component

6graNik:
Added component props to loading component.

I bumped in to situation where I need to make some logic inside loading component this props passed to Loadable.

For example:
```js
export default Loadable({
  loader: () => import('./component'),
  loading: (props) => <LoadingComponent {...props} /> 
})

class App extends React.Component {
  render() {
    return() {
      <Loadable beforeLoading={noop} afterLoading={noop}/>
    } 
  }
}

class LoadingComponent extends React.Component {
  componentDidMount() {
    this.props.beforeLoading()
  }

  componentWillUnmount() {
    this.props.afterLoading()
  }

  render() {
    <span>Loading...</span>
  }
}
```

### DIFF
--- a/__tests__/__snapshots__/test.js.snap
+++ b/__tests__/__snapshots__/test.js.snap
@@ -3,21 +3,21 @@
 exports[`delay and timeout 1`] = `
 <div>
   MyLoadingComponent 
-  {"isLoading":true,"pastDelay":false,"timedOut":false,"error":null}
+  {"isLoading":true,"pastDelay":false,"timedOut":false,"error":null,"prop":"foo"}
 </div>
 `;
 
 exports[`delay and timeout 2`] = `
 <div>
   MyLoadingComponent 
-  {"isLoading":true,"pastDelay":true,"timedOut":false,"error":null}
+  {"isLoading":true,"pastDelay":true,"timedOut":false,"error":null,"prop":"foo"}
 </div>
 `;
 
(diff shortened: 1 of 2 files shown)


6graNik:
@jamiebuilds  q

jamiebuilds:
Look at past pull requests before you open a new one

thomasjm:
@jamiebuilds the fact that so many other people have opened PRs (and AFAICT are still struggling to understand the proposed workarounds in the ongoing discussion on e.g. #88, #113) seems to indicate a lot of demand for this :P.

For me the use-case is simply to pass some props to the loading component so that the loading component can do a better "quick and dirty" render before being replaced by the real component. (I don't want to do more fancy things like "dynamically choosing the component based on the props" as #88 mentions.) You said that supporting this doesn't play well with SSR, but I can't imagine why my simple use case would cause a problem.

Maybe some of the more dynamic things people want to do wouldn't work with SSR, but I think this would be a more capable library if it just passed the props to the loader since a) not everyone wants to do fancy dynamic things and b) not everyone is using SSR. 

Maybe there could be an asterisk in the documentation saying dynamic loader tricks aren't supported but any "normal" React patterns will work fine. FWIW passing the props seems like the completely obvious way this should work (I was surprised when it wasn't) whereas overriding the Loadable `render` method is unintuitive.

jamiebuilds:
So fork it, I don't want it

PR #124, opened 2018-05-23
Fix timedOut not being reset on retry.

audiolion:
Hi 👋 

Was integrating with the new `retry` functionality introduced in #82 , really nice feature, the docs showcase using the `retry` prop when the `timedOut` prop has been set to `true`. However, the prop doesn't get reset on `retry` so the `Loading` component continues to received `timedOut: true` and never updates to the loading.

Simple fix just to default `timedOut: false` when you retry

Thanks for this great library!

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -213,7 +213,7 @@ function createLoadableComponent(loadFn, options) {
     }
 
     retry = () => {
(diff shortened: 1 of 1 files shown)


audiolion:
heya its the two month bump @jamiebuilds

PR #125, opened 2018-05-26
Don't show loading component until pastDelay === true

appleJax:
Awesome library, I'm really glad I found it! Really makes lazy-loading a breeze! 🌬️🍃 One thing I noticed is the `delay` prop does not currently work. This PR adds an extra check for `this.state.pastDelay` before showing the loading component.


blling:
Should check it in the loading component

appleJax:
@blling thanks, I guess I kinda skimmed over the README. Passing a prop to the loading component makes sense for the `error` state, but I'm not sure why the responsibility for `delay` is passed off to the user. In case someone wants to have two different loading states?

PR #132, opened 2018-06-27
Use webpack asset system instead of manually writing file

timneutkens:
This makes sure webpack its `output.path` option is respected. 

Currently, when for example a library like Next.js implements the loader it will start writing the filename where the command is executed instead of where Next.js tells build files to go (`.next`).

These implemented methods are based on `RawSource` from `webpack-sources` [link](https://github.com/webpack/webpack-sources/blob/master/lib/RawSource.js), as this library brings dependencies like `source-map` and `source-list-map`. I decided it would be beneficial to only implement the methods required instead of adding a new dependency.

I wonder if this can be made non-breaking for existing implementations. It seems like webpack doesn't respect absolute file paths in `compilation.assets`

### DIFF
--- a/src/webpack.js
+++ b/src/webpack.js
@@ -1,6 +1,4 @@
 'use strict';
-const fs = require('fs');
-const path = require('path');
 const url = require('url');
 
 function buildManifest(compiler, compilation) {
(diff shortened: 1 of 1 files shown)


timneutkens:
Going to close this as it's stale, however the described changes would still be useful for react-loadable users.

PR #134, opened 2018-07-02
Pass props to Loading component

michaltk:
There are cases when would want to render the loader with the props.
This passes the props to the loading component, so can render them in the loader.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -226,7 +226,8 @@ function createLoadableComponent(loadFn, options) {
           pastDelay: this.state.pastDelay,
           timedOut: this.state.timedOut,
           error: this.state.error,
-          retry: this.retry
+          retry: this.retry,
+          props: this.props
         });
       } else if (this.state.loaded) {
         return opts.render(this.state.loaded, this.props);


inooid:
Would love to have access to the props as well. The reason why this could be useful is whenever you render a skeleton based on a dynamic set height for a certain component, we would like the skeleton to behave the same way so we would like to have access to the height.

**Example:**
```js
// Component.js
const Component = ({ height }) => <div style={{ height }}>Awesome component</div>

// LoadableComponent.js
const Skeleton = ({ props: { height } }) => <div style={{ height }}>Some loading stuff here</div>

  loading: Skeleton,
  loader: () => import('./Component.js'),
});
```

**Current workaround:**
Basically wrap the component around another function, pass the props from there and spread them over the LoadableComponent. This is a wrapper that's not needed if we would have access to the props directly 👍 

```js
// Component.js
const Component = ({ height }) => <div style={{ height }}>Awesome component</div>

// LoadableComponent.js
const Skeleton = ({ height }) => <div style={{ height }}>Some loading stuff here</div>

const LoadableComponent = (props) => {
  const { height } = props;
  const Component = Loadable({
    loading: () => <Skeleton height={height} />,
    loader: () => import('./Component.js'),
  });

  return <Component {...props} />;
}
```

--------

⚠️ **UPDATE:** I would highly discourage you from using my recommendation above. It creates a component on the fly, which means that on re-render the React component will always be unmounted and remounted (since the references are not the same, so the reconciler thinks it's a new component every render).

inooid:
Thanks for the heads up @madiodio! This is really useful insight.

Pajn:
The current API design makes it impossible to dynamically load something like react-beautiful-dnd.
If props where passed to the loading component it could be used like this:
```
const LazyDragDropContext = Loadable({
  loader: () => import('react-beautiful-dnd').then(module => module.DragDropContext),
  loading: (props) => props.children,
})
```

But now it doesn't work to just lazy load the drag and drop behavior as I would clear the complete application until the chunk is loaded. This limitation is silly and prevents important use cases.

jarandmi:
A real life use case: I dynamic load components on our website, but i keep the menu and header visible when the new components is loading. I want to send the navigation prop to the loader, but this is not possible now. I can't pass on parents prop without this PR.

I don't see why this would mean trouble?

Pajn:
@grebenyuksv-preply I would recommend looking at `loadable-components` instead. With it you can do this without hacks. I migrated all my code and it's far simpler now.

Nikodermus:
@jamiebuilds what is your reasoning for not wanting the loader to be able to receive the props? I checked the other convos but I didn't find anything beyond "I don' want it"

Thank you for the effort on the library

kg-currenxie:
My workaround:

```
yarn add patch-package postinstall-postinstall
```
Change package.json:
```diff
 "scripts": {
+  "postinstall": "patch-package"
 }
```
Manually change `node_modules/react-loadable/lib/index.js` with the same change as this PR. Just save the file and close it.
```
yarn patch-package react-loadable
```
Done

PR #142, opened 2018-08-11
add mention about react-loadable-ts-transformer to README

stushurik:
I have created a ts transformer which adds webpack and modules props to loadable config.
I think it is a good start for typescript developers


luqimin:
this is really needed

michelalbers:
Omg, I spent days looking for this. Thank you SO MUCH!

PR #85, opened 2017-11-14
Transforming Object.assign() with Babel (fix IE11 issue)

PerrineBoissieres:
Thanks a lot for this awesome library!

We detected an issue on IE11, as Babel does not include the polyfill for `Object.assign()` in its es2015 preset. You need to include yourself the plugin to transform it, so here it is ^^

PR #144, opened 2018-08-14
Transform Object.assign with Babel in .babelrc

giannif:
Replaces `Object.assign` with an inline helper

Not sure if #85 used to work, but adding the plugin to the .babelrc applies the transform as expected

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -7,5 +7,6 @@
   ],
   "plugins": [
     "transform-class-properties",
+    "transform-object-assign",
   ],
 }


miguel-orange:
@tajo,
In GitHub it seems that tag v5.5.0 has this implemented.
I downloaded v5.5.0 from npmjs and looking into node_modules and it looks like `lib/index.js` line 107 is not extending. I don't think it's an implementation issue, I believe that libs for that tag weren't rebuilt correctly before publishing to npm.

Peroluo:
Support ie11?

PR #145, opened 2018-08-15
Option to not server-render specific components

overlookmotel:
This PR adds two options `noSsr` and `noReport`.

### `noSsr` option

`noSsr: true` tells react-loadable not to render the component on the server.

This is useful if you are doing SSR, but e.g. the component appears "below the fold" and you want to leave it to the client to render. It'll be rendered on the server as "Loading" component.

```js
Loadable({
  loader: () => import('./Bar'),
  loading: Loading,
  noSsr: true
});
```

### `noReport` option

However, you may or may not want to include the `<script>` tag for this chunk in the HTML.

Most likely you don't - you want it to be async `import()`-ed after the page has become live.

But in some cases, you might want to push the script to the client to start downloading, knowing that it's going to need it soon anyway.

The `noReport` option helps control this.

With `noReport: true`, during server-side render, the `report()` prop of `Loadable.Capture` will not be called. So then the script won't be included in the HTML.

This is the default - when `noSsr` is true, `noReport` is set to `true` too by default.

If you *do* want to put the script in the HTML, use `noSsr: true, noReport: false`. Then `report()` will be called and the script can be added to `modules` array and then added to the HTML.

But... you don't want loading these scripts to block the react app being hydrated on client. You want them to load in the background. To differentiate between modules required for hydration, and ones which are not, `report()` is called with a 2nd arg `rendered`.

### `noReport` example

So, if you're using `noSsr: true, noReport: false` in any components, you'd do something like this:

```js
// Component
Loadable({
  loader: () => import('./Bar'),
  loading: Loading,
  noSsr: true,
  noReport: false
});

// Server-side rendering
app.get('/', (req, res) => {
  // Split modules into 2 arrays - rendered and unrendered
  const renderedModules = [], unrenderedModules = [];

  const report = (moduleName, rendered) => {
    const modules = rendered ? renderedModules : unrenderedModules;
    modules.push(moduleName);
  }

  const html = ReactDOMServer.renderToString(
    <Loadable.Capture report={report}>
      <App/>
    </Loadable.Capture>
  );

  // Split bundles into two arrays - rendered and unrendered
  const renderedBundles = getBundles(stats, renderedModules);
  const unrenderedBundles = getBundles(stats, unrenderedModules);

  res.send(`
    <!doctype html>
    <html lang="en">
      <head>...</head>
      <body>
        <div id="app">${html}</div>

        <!-- Script tags for SSR-rendered components -->
        ${renderedBundles.map(bundle => {
          return `<script src="/dist/${bundle.file}"></script>`
        }).join('\n')}

        <!-- Script tag for main bundle -->
        <script src="/dist/main.js"></script>

        <!-- App has been hydrated before scripts below here execute -->

        <!-- Script tags for unrendered components -->
        ${unrenderedBundles.map(bundle => {
          return `<script src="/dist/${bundle.file}"></script>`
        }).join('\n')}

      </body>
    </html>
  `);
});
```

* The client starts loading all the scripts as soon as the page loads.
* As soon as all the top set of scripts (only those required to hydrate the react app) are loaded, the react app is hydrated.
* The bottom set of scripts continue loading in the background and will run once they've loaded.

### Next steps

This is an example API. I'm completely open to suggested changes. But I feel the functionality would be useful.

I've also not written any tests. I'd be happy to do so, but would like to get some feedback on whether this PR would be likely to be accepted and whether you'd like to see changes to API before proceeding further.

Please let me know your thoughts...

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -110,10 +110,15 @@ function createLoadableComponent(loadFn, options) {
       timeout: null,
       render: render,
       webpack: null,
-      modules: null
+      modules: null,
+      noSsr: false,
+      noReport: undefined
     },
     options
   );
+  if (opts.noReport === undefined) {
+    opts.noReport = opts.noSsr;
+  }
 
   let res = null;
 
@@ -124,14 +129,16 @@ function createLoadableComponent(loadFn, options) {
     return res.promise;
   }
 
-  ALL_INITIALIZERS.push(init);
+  if (!opts.noSsr) {
+    ALL_INITIALIZERS.push(init);
 
-  if (typeof opts.webpack === "function") {
-    READY_INITIALIZERS.push(() => {
-      if (isWebpackReady(opts.webpack)) {
-        return init();
-      }
-    });
+    if (typeof opts.webpack === "function") {
+      READY_INITIALIZERS.push(() => {
+        if (isWebpackReady(opts.webpack)) {
+          return init();
+        }
+      });
+    }
   }
 
   return class LoadableComponent extends React.Component {
@@ -164,13 +171,13 @@ function createLoadableComponent(loadFn, options) {
     }
 
     _loadModule() {
-      if (this.context.loadable && Array.isArray(opts.modules)) {
+      if (this.context.loadable && Array.isArray(opts.modules) && !opts.noReport) {
         opts.modules.forEach(moduleName => {
-          this.context.loadable.report(moduleName);
+          this.context.loadable.report(moduleName, !opts.noSsr);
         });
       }
 
-      if (!res.loading) {
+      if (!res.loading || opts.noSsr) {
         return;
       }
 


Lysholm:
Thanks, this was exactly what i was looking for! I think I'd prefer the options to be
`ssr = true: boolean` and `report = ssr: boolean`. Then the API is be similar to [next/dynamic](https://github.com/zeit/next.js#3-with-no-ssr) and avoids the unnecessary negation.

jamiebuilds:
I'd rather have these flipped as:

```js
Loadable({
  loader: () => import('./Bar'),
  loading: Loading,
  ssr: false,
  report: false,
});
```

overlookmotel:
I'm afraid it's been a while, so I've moved on, and built my [own thing](https://www.npmjs.com/package/react-lazy-ssr) to accomplish the same thing.

I'm going to close this PR as I'm sure it has bugs in it and I don't want to be responsible!

PR #154, opened 2018-09-07
Remove header with expletive

jamescun:
Expletives in this project could make it harder to use in a corporate environment or in training (often of young people).

I do not have the source file for the header image, so unfortunately can only remove the header image itself.


kaelanfouwels:
+1. Please think of the children when adding such expletive and vulgar language to a repository.

jamiebuilds:
Corporations can eat my ass

billinghamj:
😢

PR #161, opened 2018-09-26
Support passing children to the loading component

randing89:
In our use case, rendering the `children` while displaying the loading component is critical for performance reason.
This PR adds the support in a backward-compatible way.

### DIFF
--- a/README.md
+++ b/README.md
@@ -747,6 +747,17 @@ Loadable({
 });
 ```
 
+#### `opts.children`
+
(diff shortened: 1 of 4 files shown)


stevemao:
Why not creating different `LoadingComponent`s which is a wrapper of master `LoadingComponent` with different children?

randing89:
@stevemao You still need some way to get the children from the consumer.

jamiebuilds:
You can do this already by setting your own [`render()` method](https://github.com/jamiebuilds/react-loadable#optsrender)

PR #163, opened 2018-10-07
Support emitting assets via webpack build

jahredhope:
The webpack loader currently outputs a file directly to dist, which can then be picked up by the consumer's server render.
It's possible to pass in custom file systems to webpack, such as [memory-fs](https://github.com/webpack/memory-fs).

This change allows the loader to optionally output assets directly into the Webpack output.
This change is non-breaking, and opted into by passing `emitAssets: true`.

I believe it may make sense to have `emitAssets` be the default behaviour, but that could break consumer builds.

# Examples

**Write to disk**
``` js
    new ReactLoadablePlugin({
      writeToDisk: true, // or undefined
      filename:  path.resolve(__dirname, 'example', 'dist', 'react-loadable.json'),
    })
```

**Emit to Webpack assets**
``` js
    new ReactLoadablePlugin({
      writeToDisk: false,
      emitAssets: true,
      filename: 'react-loadable.json',
    })
```
**Example output of emitting assets**
Example webpack output including a react-loadable.json file.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,8 @@
     "prepublish": "yarn build"
   },
   "dependencies": {
-    "prop-types": "^15.5.0"
+    "prop-types": "^15.5.0",
+    "webpack-sources": "^1.3.0"
   },
   "devDependencies": {
     "babel-cli": "^6.24.1",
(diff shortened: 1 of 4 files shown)


jamiebuilds:
Can it just emitAssets all the time?

jamiebuilds:
I think another PR has handled this

PR #166, opened 2018-10-24
Improve SSR integration documentation

followdarko:
SSR integration documentation not match with real example. It's a bit confusing when you trying to do it. This PR makes it more clear


ryardley:
Usually the main bundle and react render code is at the bottom of the file anyway why add an extra layer of indirection?

PR #169, opened 2018-11-07
Add getWrappedInstance method for LoadableComponent

zachary-zhuangli:
Thanks a lot for this awesome library!

In many use cases, we need the getWrappedInstance method to get the wrapped Instance, like react-redux and react-intl do, and here it is.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -229,6 +229,10 @@ function createLoadableComponent(loadFn, options) {
       this._loadModule();
     };
 
+    getWrappedInstance = () => {
+      return this._wrappedInstance;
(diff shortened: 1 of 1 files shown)


pi-kei:
What should be done to make this PR be merged?

lz19951209:
What should be done to make this PR be merged? many projects need this method...plz

qq281113270:
> Thanks a lot for this awesome library!
> 
> In many use cases, we need the getWrappedInstance method to get the wrapped Instance, like react-redux and react-intl do, and here it is.

PR #170, opened 2018-11-12
ignore undefined key from manifest

hardfist:
Change-Id: I684bb8f54310ab3f9cccd5927ae352bb84bc2a0e

### DIFF
--- a/src/webpack.js
+++ b/src/webpack.js
@@ -13,16 +13,17 @@ function buildManifest(compiler, compilation) {
         let id = module.id;
         let name = typeof module.libIdent === 'function' ? module.libIdent({ context }) : null;
         let publicPath = url.resolve(compilation.outputOptions.publicPath || '', file);
-        
+
         let currentModule = module;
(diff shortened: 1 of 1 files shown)


hardfist:
`currentModule.rawRequest` maybe `undefined` which makes manifest containes a 	`undefiend` key, which seems doesn't make sense.

intelcoder:
Do you happen to know how to prevent undefined?
Some of component is used in server side but it is not fetching on server side and fetches when client side takes over. This causes slow page load.
I have already asked many times on Stackoverflow and used bounty hunter feature. However, no one could answer that.

beHitesh:
Hi, I'm facing a similar issue. Has anyone figured a solution for this?

bertho-zero:
@beHitesh By using @loadable/components, it's a little less closed than react-loadable...

qq281113270:
> Change-Id: I684bb8f54310ab3f9cccd5927ae352bb84bc2a0e

PR #171, opened 2018-11-28
use compiler.outputFileSystem

yujiaze:
use `compiler.outputFileSystem` instead of native `fs` module

we prefer to use `memory-fs` in dev mode ( eg: `webpack-dev-middleware`), current implements in this plugin using native `fs` module may produce conflicts in code.

### DIFF
--- a/src/webpack.js
+++ b/src/webpack.js
@@ -1,5 +1,4 @@
 'use strict';
-const fs = require('fs');
 const path = require('path');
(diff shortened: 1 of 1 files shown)


Schachte:
We should merge this as webpack will default to fs. @jamiebuilds

jamiebuilds:
I think another PR has handled this

PR #175, opened 2018-12-26
Fix render parameter's property

MartinYounghoonKim:
Fix 'namedExport' to 'default'


jamiebuilds:
This is to demonstrate what to do when you are exporting your component with a named export.

PR #179, opened 2019-02-11
should throw error after catch

clinyong:


clinyong:
I am sorry that haven't run test locally. I want the error can be throw outside so that the global handler(using the [unhandledrejection](https://developer.mozilla.org/en-US/docs/Web/Events/unhandledrejection), although the compatibilitySection is bad.) can  catch it. But since it will cause jest test fail, I have to close the PR.

PR #192, opened 2019-07-15
Add “BBC News” to users

dr3:


UmerZafar990:
Kindly add [**ilaan**](https://www.ilaan.com/) to the users list.

PR #195, opened 2019-08-12
Remove use of componentWillMount

dr3:
React 16.9.0 [deprecated](https://reactjs.org/blog/2019/08/08/react-v16.9.0.html#renaming-unsafe-lifecycle-methods) `componentWillMount` so it now throws a warning when used. 

As far as I can see this should be a simple swap to `componentDidMount`

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -158,7 +158,7 @@ function createLoadableComponent(loadFn, options) {
       return init();
     }
 
-    componentWillMount() {
+    componentDidMount() {
       this._mounted = true;
       this._loadModule();
     }
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-loadable",
-  "version": "5.5.0",
+  "version": "5.5.1",
   "description": "A higher order component for loading components with promises",
   "main": "lib/index.js",
   "author": "James Kyle <me@thejameskyle.com>",


dr3:
@jamiebuilds This looks like it passes on node 6,8 & 10 and 4 fails for an unrelated reason. Ready for code review :)

bertho-zero:
@dr3 This project is dead.

dr3:
@bertho-zero And not accepting any contributions at all? Seems a shame, works well for those with SSR until Lazy supports it

prevolorio:
ya it looks like this library has been replaced by React Suspense

dr3:
@prevolorio It looks that way, which is a shame as Suspense doesn't support SSR yet

pierreneter:
Should call ` this._loadModule();` in `constructor`.

pierreneter:
@jamiebuilds pls review this PR, I wonder why this project has no issues tab? It's easier to find out what's wrong. And, we can stop support for Node 4.

src/index.js
```diff
@@ -158,7 +158,7 @@ function createLoadableComponent(loadFn, options) {
       return init();
     }
 
-    componentWillMount() {
+    componentDidMount() {
       this._mounted = true;
       this._loadModule();
```
pierreneter: should call `_loadModule` in `constructor`

pierreneter: And handle it:
https://github.com/jamiebuilds/react-loadable/pull/195/files#diff-1fdf421c05c1140f6d71444ea2b27638R198-R202

dr3: Do you recommend where to split the logic? obs cant call setState in the constructor, but _loadModule is also used in the retry method

TSMMark:
What will it take to get this merged?

bertho-zero:
Still alive and maintained: https://github.com/smooth-code/loadable-components

vignestion:
Any update on this, will this PR get merged anytime soon?

ankibalyan:
Atleast this depricated issue should be resolved before suspense supports SSR.

leantide:
@ankibalyan One would certainly think this 1 PR would go in. One would think a warning would be posted on the project entry page that this is unmaintained. One would think that this project would be archived. One would think a referral link to https://github.com/smooth-code/loadable-components would be in order. One would think that github would have some mitigating measurement of new vs old stars. But alas...

But thankfully migrating over to https://github.com/smooth-code/loadable-components is probably quite painless for most. At least it was for me.

jasonnathan:
Build is failing for Node 4, perhaps the author wants to still support Node 4?

dr3:
@jasonnathan the `master` branch also fails on node 4

cdeutsch:
> Still alive and maintained: https://github.com/smooth-code/loadable-components

That doesn't have the same features or API, unless I'm missing something?

src/index.js
cdeutsch: Slightly ugly, but you could make an internal `_ setState`

```js
_setState(state) {
  if (this._mounted) {
    this.setState(state);
  } else {
    this.state = state;
  }
}
```

cdeutsch: Actually I tried this and React will still warn about causing SideEffects.

So, I don't think `_loadModule` (as-is) should be called in the `constructor` as suggested.

If you want to try it, it's here and on NPM (but I wouldn't recommend it)
https://github.com/axiomhq/react-loadable

cdeutsch: Probably better to rename it to `UNSAFE_ componentWillMount` and call it a day

cdeutsch: I ended up trying out this change (using `componentDidMount`) in `@axiomhq/react-loadable@5.5.2`.

It also has side effect warnings.

It looks like the new React API is just too different to make this easily work.

image

yanlee26:
Shame

bertho-zero:
@cdeutsch There are the same features with https://github.com/smooth-code/loadable-components

dr3:
Yeah we were considering forking this repo under the `bbc` organisation, however were now moving to loadable-components, similar features with actually a fair bit more stuff. 

Happy to merge this, if allowed, if anyone wants it though

mkinfrared:
> Yeah we were considering forking this repo under the `bbc` organisation, however were now moving to loadable-components, similar features with actually a fair bit more stuff.
> 
> Happy to merge this, if allowed, if anyone wants it though

It has typescript support unlike loadable-components

mbasakCinedigmInd:
still dont know how to solve it. can you please help me?

fangyinghua:

PR #201, opened 2019-09-10
Rename unsafe lifecycle hooks

chaudharykiran:
1. Unsafe lifecycles hook will not work from React v17.x without prefix with UNSAFE_
2. This will fix the warning message in React v16.9.0


qq281113270:
> 1. Unsafe lifecycles hook will not work from React v17.x without prefix with UNSAFE_
> 2. This will fix the warning message in React v16.9.0

PR #214, opened 2019-10-17
Add FAQ question to README

TylerSustare:
I had never used import() as a function before using react-loadable. I ran into this and it took me a little bit to fix. Hopefully, someone else can benefit from it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1233,3 +1233,19 @@ res.send(`
   </html>
 `);
 ```
+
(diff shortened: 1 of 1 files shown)


TylerSustare:
I'm not quite sure why the Travis build would fail for this. Is there something I did wrong?

jamiebuilds:
Probably would be better just to add a warning to the code.

PR #213, opened 2019-10-14
Added `UNSAFE_` prefix to componentWillMount()

smrsan76:
Hello dear maintainers,

Please merge and publish this new code to omit the warning message. :pray:

Thanks

PR #217, opened 2019-11-11
Added `UNSAFE_` prefix to componentWillMount() (2)

ramirost:
Is the PR #213 with the version bump of react dev dependency to fix tests


ramirost:
PR #126 already solves the same problem and tests are passing

PR #226, opened 2020-03-09
Add support for adding a displayName property for the returned Loadable component

backyardcoder:
The primary reason to introduce this change is to provide a way to aid the testing of components that consumes a Loadable component. 
For example:

```javascript
// In source code file:
const MyComponent = Loadable({
  displayName: 'MyComponent', // same as the import name and guaranteed to be unique in the "sut"
  loader: () => ...,
  loading: () => ...
});

// In test file
import MyComponent from '../my-path';

fakeLoader.withArgs({
   displayName: 'MyComponent',
   loader: sinon.match.func,
   loading: sinon.match.func
}).returns(MyComponent);
```

With the help of `displayName`, one can use a package like `proxyquire` to set up the correct fake loader stub to return expected components.

### DIFF
--- a/README.md
+++ b/README.md
@@ -779,6 +779,17 @@ Loadable({
 });
 ```
 
+#### `opts.displayName`
+
+Display name for the returned component. Defaults to `LoadableComponent`.
+
+```js
+Loadable({
+  displayName: 'MyLoadableComponent',
+  loader: () => import('./Bar'),
+});
+```
+
 #### `opts.delay`
 
 Time to wait (in milliseconds) before passing
--- a/__tests__/test.js
+++ b/__tests__/test.js
@@ -201,6 +201,23 @@ test('loadable map error', async () => {
   expect(component.toJSON()).toMatchSnapshot(); // success
 });
 
+test('display name', async () => {
+  let LoadableMyComponent = Loadable({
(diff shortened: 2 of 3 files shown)


backyardcoder:
What do you think about adding this support, @jamiebuilds @tajo?

StefanMcCready:
It would be super cool to have this as part of the project. For a certain project, I am relying on the component displayName to help pick out certain components from children. Since we have moved to use react-loadable we have not been able to do this.

qq281113270:
> The primary reason to introduce this change is to provide a way to aid the testing of components that consumes a Loadable component. For example:
> 
> ```js
> // In source code file:
> const MyComponent = Loadable({
>   displayName: 'MyComponent', // same as the import name and guaranteed to be unique in the "sut"
>   loader: () => ...,
>   loading: () => ...
> });
> 
> // In test file
> import MyComponent from '../my-path';
> 
> fakeLoader.withArgs({
>    displayName: 'MyComponent',
>    loader: sinon.match.func,
>    loading: sinon.match.func
> }).returns(MyComponent);
> ```
> 
> With the help of `displayName`, one can use a package like `proxyquire` to set up the correct fake loader stub to return expected components.

PR #230, opened 2020-05-03
Found a typo

frankda:
Dear Maintainer

I just found a typo. In example code, it delays 0.3s while in text it shows `300ms`.

Best Regards,

Frank

### DIFF
--- a/README.md
+++ b/README.md
@@ -294,7 +294,7 @@ function Loading(props) {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


jamiebuilds:
It does _default_ to `200ms`: https://github.com/jamiebuilds/react-loadable/blob/master/src/index.js#L109

The following example is showing how to override it to `300`

PR #232, opened 2020-05-18
feat: post props to loading Component

RobotLivermore:
I met a situation that the loading part could not met my need. The situation is like that:
```
const Loading = (props) => {
    if ( props.condition) {
        return <A {...props} />;
    } else {
        return <B {...props} />;
    }
}
```
Maybe leave a channel to post props is needed.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -245,7 +245,8 @@ function createLoadableComponent(loadFn, options) {
           pastDelay: this.state.pastDelay,
           timedOut: this.state.timedOut,
           error: this.state.error,
-          retry: this.retry
+          retry: this.retry,
+          initialProps: this.props
         });
       } else if (this.state.loaded) {
         return opts.render(this.state.loaded, this.props);
--- a/src/index.js
+++ b/src/index.js
@@ -246,7 +246,7 @@ function createLoadableComponent(loadFn, options) {
           timedOut: this.state.timedOut,
           error: this.state.error,
           retry: this.retry,
-          initialProps: this.props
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -245,7 +245,8 @@ function createLoadableComponent(loadFn, options) {
           pastDelay: this.state.pastDelay,
           timedOut: this.state.timedOut,
           error: this.state.error,
-          retry: this.retry
+          retry: this.retry,
+          initialProps: this.props
```
EL-shadow: Just looking for the same!

EL-shadow:
@jamiebuilds what do you think about this? It would be great to merge it.

M-OC:
+1 on this.

src/index.js
vepasto: Why to name it as `initialProps`? I would name it as `props` because the value will change whenever props changes.

RobotLivermore: `props.props` looks something strange. I think you are right  that `initialProps` is not so clear.

RobotLivermore: I have changed it now.

qq281113270:
> I met a situation that the loading part could not met my need. The situation is like that:
> 
> ```
> const Loading = (props) => {
>     if ( props.condition) {
>         return <A {...props} />;
>     } else {
>         return <B {...props} />;
>     }
> }
> ```
> 
> Maybe leave a channel to post props is needed.

fangyinghua:

PR #233, opened 2020-06-20
Add Kapwing to readme

justswim:
We use react-loadable at Kapwing in order to load many of our client side components. It has improved our bundle size as well as our Google scores a lot! We'd love if we could be included in the example list. Thanks!


qq281113270:
> We use react-loadable at Kapwing in order to load many of our client side components. It has improved our bundle size as well as our Google scores a lot! We'd love if we could be included in the example list. Thanks!

PR #239, opened 2020-09-07
Fix moduleName conflict by return module ID

GitaiQAQ:


qq281113270:
> _ No description provided. _

PR #241, opened 2020-12-23
Add Wave Digital to the list of users

t49tran:
First, thanks for this awesome library. We have been using it for all of our projects that require SSR and Code Splitting. 

Can our company Wave Digital  be added to the list of users ?

Thanks and happy new year from Australia.


qq281113270:
> First, thanks for this awesome library. We have been using it for all of our projects that require SSR and Code Splitting.
> 
> Can our company Wave Digital be added to the list of users ?
> 
> Thanks and happy new year from Australia.

PR #242, opened 2021-01-06
Fix DeprecationWarning: Chunk.forEachModule

JasonKleban:
DeprecationWarning: Chunk.forEachModule: Use for(const module of chunk.modulesIterable) instead


qq281113270:
> DeprecationWarning: Chunk.forEachModule: Use for(const module of chunk.modulesIterable) instead

PR #245, opened 2021-07-15
fix example

yiminghe:
upgrade webpack


qq281113270:
> upgrade webpack

qq281113270:
> upgrade webpack

PR #247, opened 2021-10-07
New `ssr` option

trhinehart-godaddy:
Support optional `ssr` option to allow `Loadable.preloadAll(true)` to only preload modules that are needed for SSR.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -3,6 +3,7 @@ const React = require("react");
 const PropTypes = require("prop-types");
 
 const ALL_INITIALIZERS = [];
+const SSR_INITIALIZERS = [];
 const READY_INITIALIZERS = [];
 
 function isWebpackReady(getModuleIds) {
@@ -110,7 +111,8 @@ function createLoadableComponent(loadFn, options) {
       timeout: null,
       render: render,
       webpack: null,
-      modules: null
+      modules: null,
+      ssr: true
     },
     options
   );
@@ -125,6 +127,10 @@ function createLoadableComponent(loadFn, options) {
   }
 
   ALL_INITIALIZERS.push(init);
+  
+  if (opts.ssr) {
+    SSR_INITIALIZERS.push(init);
+  }
 
   if (typeof opts.webpack === "function") {
     READY_INITIALIZERS.push(() => {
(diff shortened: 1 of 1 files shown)


dobesv:
Our use case: we only have a small subset of components that are visible to SSR (e.g. publically available content).  In our e2e tests, `preloadAll()` takes over 3 minutes as we have a lot of components.  It would be really great if SSR could just load the components needed for that specific page being rendered.

dobesv:
Although, I guess this solution requires us to just statically annotate the loadable components to indicate which ones to preload, whereas I would prefer a solution where there can be async rendering that loads the modules needed to render the page and nothing extra.

trhinehart-godaddy:
> I would prefer a solution where there can be async rendering that loads the modules needed to render the page and nothing extra.

I have a solution in mind for that, however at this rate of feedback from @jamiebuilds I'm not sure if this is the right library to contribute to.

dobesv:
Yeah, I'm going to experiment with the other loadable components library and see if it helps.

ygs-code:
After I upgraded webpack 4 to wepack5 。The react-loadable plugin failed chunk.forEachModule is not a function.
code：

function buildManifest(compiler, compilation) {
  var context = compiler.options.context;
  var manifest = {};

  compilation.chunks.forEach(function (chunk) {
    chunk.files.forEach(function (file) {
      chunk.forEachModule(function (module) {
        var id = module.id;
        var name = typeof module.libIdent === 'function' ? module.libIdent({ context: context }) : null;
        var publicPath = url.resolve(compilation.outputOptions.publicPath || '', file);

        var currentModule = module;
        if (module.constructor.name === 'ConcatenatedModule') {
          currentModule = module.rootModule;
        }
        if (!manifest[currentModule.rawRequest]) {
          manifest[currentModule.rawRequest] = [];
        }

        manifest[currentModule.rawRequest].push({ id: id, name: name, file: file, publicPath: publicPath });
      });
    });
  });

  return manifest;
}

How can I fix this bug and make it work？

qq281113270:
> Support optional `ssr` option to allow `Loadable.preloadAll(true)` to only preload modules that are needed for SSR.

fangyinghua:

